PR #1, opened 2022-06-25
Send all typedData to to the /sign iframe

sam0110:
Sends all of the typedData to the /sign iframe so that it can read the types.

### DIFF
--- a/sdk/src/account.ts
+++ b/sdk/src/account.ts
@@ -206,7 +206,7 @@ export class Account extends Provider implements AccountInterface {
       `${this.url}/sign?${qs.stringify({
         id,
         origin: window.origin,
-        message: JSON.stringify(typedData.message),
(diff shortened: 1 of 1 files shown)


sdk/src/account.ts
```diff
@@ -206,7 +206,7 @@ export class Account extends Provider implements AccountInterface {
       `${this.url}/sign?${qs.stringify({
         id,
         origin: window.origin,
-        message: JSON.stringify(typedData.message),
+        message: JSON.stringify(typedData),
```
tarrencev, line 209:
```suggestion
        typedData: JSON.stringify(typedData),
```

tarrencev (approved):

sdk/src/account.ts
sam0110: Fixed!

PR #41, opened 2022-12-13
feat(keychain): save deployment tx

broody:


keychain/src/methods/register.ts
```diff
@@ -39,11 +34,23 @@ const register =
       "0",
     );
 
-    const controller = new Controller(keypair, address, credentialId);
```
tarrencev, line 42: i think we should still keep this, since a controller can exist without being deployed. in some cases, the user will deploy it themselves

keychain/src/utils/account.ts
```diff
@@ -42,12 +44,38 @@ class Account extends BaseAccount {
...
+            case "ACCEPTED_ON_L1":
+            case "ACCEPTED_ON_L2":
+              this.deploying = false;
+              this.deployed = true;
+              break;
+            case "RECEIVED":
+            case "PENDING":
+              this.deploying = true;
+              this.deployed = false;
+              throw new Error("Deploying");
```
tarrencev, line 64: can we return here and below instead of throwing?

broody: Yep, just turned it into a method to keep it sync cleaner and easier to call from frontend too I think

keychain/src/methods/register.ts
```diff
@@ -43,7 +43,14 @@ const register =
     controller.store();
     Storage.set(selectors["0.0.3"].active(), address);
 
-    return { address, deviceKey };
+    return { address, deviceKey, keypair };
   };
 
-export default register;
+const saveDeploy = (address: string, hash: string) => {
+  Storage.set(
```
tarrencev, line 50: this should be `update` to merge with the existing object

keychain/src/methods/register.ts
```diff
@@ -43,7 +43,14 @@ const register =
     controller.store();
     Storage.set(selectors["0.0.3"].active(), address);
 
-    return { address, deviceKey };
+    return { address, deviceKey, keypair };
   };
 
-export default register;
+const saveDeploy = (address: string, hash: string) => {
```
tarrencev, line 49: we will need to make sure the caller is authorized (i.e. an admin) like:

https://github.com/cartridge-gg/cartridge/blob/main/keychain/src/methods/sessions.ts#L34

tarrencev, line 49: we can add this to the controller interface here:

https://github.com/cartridge-gg/cartridge/blob/main/controller/src/types.ts#L38

PR #81, opened 2023-01-20
feat(keychain): transactions screen

Larkooo:


keychain/src/components/Transaction.tsx
```diff
@@ -59,15 +61,19 @@ export const Transaction = ({
...
-        <HStack color="gray.200" spacing="5px">
-          <StarknetIcon boxSize="14px" />
-          <Text color="inherit" fontSize="13px">
-            {chainId === constants.StarknetChainId.MAINNET
-              ? "Mainnet"
-              : "Testnet"}
-          </Text>
-        </HStack>
-        <Divider orientation="vertical" bgColor="gray.500" h="30px" />
+        {showChainId && (
```
tarrencev, line 64: how come this is conditional?

tarrencev (approved):

keychain/src/components/Transaction.tsx
tarrencev: Where else are we using it? Shouldn't we always show the chain id?

Larkooo: The transactions page doesn't show the chain ID. It's at the top. All transactions are the same chain. So no need to display the chain ID again (from calvins designs at least)

PR #88, opened 2023-01-24
fix: sync issues

broody:
@tarrencev in the case where we deploy, I think we need to poll for contract after signup otherwise user is stuck in `COUNTERFACTUAL` unless try start triggering actions. Or maybe only poll if deployAccount api is called?

Also, `PENDING_REGISTRATION` gets overridden by `DEPLOYED` after login. Tho with this fix, device registration isn't getting triggered still

### DIFF
--- a/keychain/src/utils/account.ts
+++ b/keychain/src/utils/account.ts
@@ -152,6 +152,10 @@ class Account extends BaseAccount {
         }
       }
 
+      if (this.status === Status.PENDING_REGISTER) {
+        return;
+      }
+
       const classHash = await this.rpc.getClassHashAt(this.address, "pending");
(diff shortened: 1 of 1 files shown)


keychain/src/utils/account.ts
```diff
@@ -117,6 +117,8 @@ class Account extends BaseAccount {
           Storage.update(this.selector, {
             status: Status.COUNTERFACTUAL,
           });
+          // async deployed, poll for contract
+          setTimeout(() => this.sync(), 5000);
```
tarrencev, line 121: Hmm I'm not sure I understand the specific flow where this is necessary? Is it if the deploy comes from another application?

PR #157, opened 2023-07-31
Setup color mode based on figma variables and add Button story

JunichiSugiura:
## Objectives

- Setup color mode with Chakra UI theme (`dark | light`) based on figma variables
- Add `Button` story

Screenshot 2023-08-01 at 15 03 34


JunichiSugiura:
:hidethepain:

- [How to fix "Expression produces a union type that is too complex to represent. ts(2590)"?](https://stackoverflow.com/questions/74847053/how-to-fix-expression-produces-a-union-type-that-is-too-complex-to-represent-t)

JunichiSugiura:
It compiled after removing `"jsxImportSource": "@emotion/react"` from `tsconfig.json` but now I got a runtime error.

Screenshot 2023-08-01 at 18 37 32

PR #228, opened 2023-12-21
Use RPC for estimate bulk fee

broody:


JunichiSugiura:
I'm leave a reference here. 

`status` -> `execution_status`

Ref: https://docs.starknet.io/documentation/architecture_and_concepts/Network_Architecture/transaction-life-cycle/#transaction_receipt

Last night when I tested this flow by changing `@deployment/` status to `REGISTERED`, the [last return of estimateInvokeFee method](https://github.com/cartridge-gg/cartridge/blob/main/packages/keychain/src/utils/account.ts#L352) still throw error.

Is there a chance that the sync patch would fix it also? @broody

broody:
I recall deployment status goes like this DEPLOYING -> DEPLOYED -> PENDING_REGISTER -> REGISTERED. I think the estimate fee is failing because your keypair is not yet registered. You can try setting it to DEPLOYED, and then on the next execute it will bundle a registration txn.

But this is still an issue for me because my webauthn is registered to x.cartridge.gg, and I can't test it with any of the preview branches. In the past, we could just deploy a new account using a preview branch, but deployment isn't working anymore (I think caigo needs to be updated).

PR #240, opened 2024-03-07
Setup ui-next package with tailwind and shadcn-ui

JunichiSugiura:
Setup `@cartridge/ui-next` package to explorer shadcn-ui.

This PR is not a high priority currently but to experiment.

close: INT-511


ponderingdemocritus:
Great path forward that will accelerate all downstream apps building off this library. Setup looks good. 

Lets map a feature parity path with the current UI. The actual library shouldn't take more than a day or so to build.

PR #241, opened 2024-03-11
Account SDK integration example via wasm-pack

JunichiSugiura:
This is an experiment PR to integrate account-sdk crate via wasm-bindgen

### To run

```sh
# Post install script generates `@cartridge/account-sdk` package
pnpm i

# Serve example page
pnpm example:account-sdk dev
```

close: INT-510


JunichiSugiura:
Run `wasm-pack build --web` without removing tokio dependency from `account-sdk` crate

Screenshot 2024-03-11 at 19 24 13

JunichiSugiura:
After remove tokio...

abi json needs to be generated pre-build

Screenshot 2024-03-11 at 19 26 11

piniom:
Hey @JunichiSugiura since you include account_sdk as a dependency using local path, you can run `make` in the root of the `cairo-webauthn` project. After this the thing should compile fine. lmk if it helps.

 I'm thinking about an elegant solution to the problem, so that the cairo-webauth/account_sdk crate can be simply included as a "remote" (github/crates.io) dependency.

broody:
> Hey @JunichiSugiura since you include account_sdk as a dependency using local path, you can run `make` in the root of the `cairo-webauthn` project. After this the thing should compile fine. lmk if it helps.
> 
> I'm thinking about an elegant solution to the problem, so that the cairo-webauth/account_sdk crate can be simply included as a "remote" (github/crates.io) dependency.

Ah seems like still an issue, I just hardcoded the an absolute path for now as a workaround. A simple include dep would be great.

JunichiSugiura:
> Hey @JunichiSugiura since you include account_sdk as a dependency using local path, you can run `make` in the root of the `cairo-webauthn` project. After this the thing should compile fine. lmk if it helps.
> 
> I'm thinking about an elegant solution to the problem, so that the cairo-webauth/account_sdk crate can be simply included as a "remote" (github/crates.io) dependency.

@piniom Yes, I tried `make` command but didn't work. Broody's hack works.

JunichiSugiura:
Screenshot 2024-03-12 at 15 54 27

JunichiSugiura:
Keychain build fails most likely due to the whitelisting on alchemy endpoints

piniom:
Hey @JunichiSugiura @broody ,
I think I've managed to streamline the inclusion of the account_sdk as a dependecy.
You can test it using the commit in this  [PR](https://github.com/cartridge-gg/cairo-webauthn/pull/44/commits):
```toml
account-sdk = { git = "https://github.com/cartridge-gg/cairo-webauthn.git", rev="0c7e87e206310a59dec3faf2b3acdf84fd469c8b" }
# account-sdk = { path = "../../../cairo-webauthn/crates/account_sdk/" }
```
The only caveat is that you need a nigtly build of rust due to one crucial feature that I've needed to make this work. 
You also need to have [scarb](https://docs.swmansion.com/scarb/) installed. 
Lmk if it works for you

JunichiSugiura:
@piniom Thanks for the [PR](https://github.com/cartridge-gg/cairo-webauthn/pull/44/commits) !
I confirm that your pre-build approach worked with this branch.

Since the approach requires nightly toolchain, I suggest adding toolchain file so the user don't have to manually switch.
Ref: [The rustup book - The toolchain file](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file)

PR #244, opened 2024-03-15
Upgrade starknetjs to v6

JunichiSugiura:
Revival of #223 

Remaining migrations are commented inside the code.
Grep for `#244`.


JunichiSugiura:
It's a bit annoying that vercel shortens address when branch name(preview url) is too long.
That's why controller never connects to keychain on this PR.

Screenshot 2024-03-22 at 15 49 37

JunichiSugiura:
Session does not seem to be stored.

Screenshot 2024-03-22 at 16 37 07

packages/keychain/src/utils/account.ts
```diff
@@ -287,90 +303,102 @@ class Account extends BaseAccount {
...
+            // TODO(#244): Set resource bounds
+            resourceBounds: {
+              l1_gas: {
+                max_amount: "",
+                max_price_per_unit: "",
+              },
+              l2_gas: {
+                max_amount: "",
+                max_price_per_unit: "",
+              },
```
JunichiSugiura, line 357: Not sure what to put. and why `overall_fee` and `gas_consumed` field still exist?

broody:
Great stuff! With this we can successfully deploy controller and transfer ETH in starknet-react-example (running local katana). I'm debugging an issue with sign message but I think we can merge this and fix that and address other TODOs in later PRs.

JunichiSugiura:
@broody Thanks for the debugging and sepolia migration !

PR #246, opened 2024-03-27
Update starknet-react example with dojo spawn-and-move

broody:


broody:
For local devnet, requires at least dojo 0.6.0-alpha.9. Have to manually deploy spawn-and-move and grant writer for now till we can configure it as part of genesis config.

Testing on sepolia should just work.

PR #254, opened 2024-04-08
Simplify account sync logic

broody:


packages/keychain/src/utils/account.ts
```diff
@@ -133,78 +140,36 @@ class Account extends BaseAccount {
...
-            this.waitingForDeploy = true;
+      switch (this.status) {
+        case Status.DEPLOYING:
+        case Status.COUNTERFACTUAL:
+          const hash = await this.getDeploymentTxn();
+          const receipt = await this.rpc.getTransactionReceipt(hash);
+          if (receipt.isRejected()) {
+            // TODO: Handle redeployment
+            this.status = Status.REJECTED;
+            return;
```
JunichiSugiura, line 151: It's usually a good practice to wrap each case in expression so each const is scoped.

```
case XXX: {
  const hash = await this.getDeploymentTxn();
  // ...
  break;
}

PR #262, opened 2024-04-11
Specify packages as ESM and extract common TS configs

JunichiSugiura:
- Specify connector + controller as ES module
- Configure TS as `Bundler` module
- Move common ts config to `tsconfig.base.json` and `tsconfig.react.json`


packages/keychain/tsconfig.json
```diff
@@ -1,37 +1,23 @@
...
-    "moduleResolution": "node",
-    "resolveJsonModule": true,
-    "isolatedModules": true,
-    "downlevelIteration": true,
-    "noEmit": true,
     "types": [
       "@emotion/react/types/css-prop"
-    ]
+    ],
+    // TODO(#262): Enable these options (too many errors for this scope)
```
JunichiSugiura, line 9: A bit too much errors to fix in this scope but should be fixed asap. A lot of type errors has been surpressed.

PR #273, opened 2024-04-23
Expose logout method via controller and connector

JunichiSugiura:


packages/connector/src/index.ts
```diff
@@ -83,6 +83,10 @@ class ControllerConnector extends Connector {
     return this.controller.disconnect();
   }
 
+  logout() {
```
tarrencev, line 86: does this allow any app to log a user out of their controller?

JunichiSugiura: Yes, was it intentionally not exposed for some reason?

JunichiSugiura: what about not directly disconnect right away but rather shows promot to the user much like how signing works?

tarrencev: we can allow an app to logout of that particular app, i.e. remove the connection / delete a session, but we cant allow any application to logout and delete the entires users local state. so this was intentionally not exposed originally

PR #338, opened 2024-06-05
Fix Cargo workspace 

kariy:
- **raise account-wasm crate one level up**
- **add to workspace**
- bumped `account_sdk` on the wasm crate to use the latest rev

was trying to fix error thrown from the WASM crate not being properly show in JS. So, fixed the workspace and make sure to use `JsError` instead of `JsValue` when returning an error from Rust.

NOTE: need to rebuild the wasm package again. otherwise the error changes wont take effect.


packages/account-wasm/tests/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+    let _ = account.sign(&challenge_bytes).await.unwrap();
+
+    // let args = VerifyWebauthnSignerArgs::from_response(
+    //     origin.to_string(),
+    //     challenge_bytes.to_vec(),
+    //     assertion,
+    // );
+
+    // let client_data = String::from_utf8(args.client_data_json).unwrap();
+    // console::log_1(&client_data.into());
```
kariy, line 28: the `VerifyWebauthnSignerArgs` seems to have been removed on the newer `account_sdk` so commented this out.

PR #346, opened 2024-06-07
Improve policy list scrolling

kariy:
Before 🤢👎:

https://github.com/cartridge-gg/controller/assets/26515232/a5070065-8626-4863-8109-9351a98720d7

After 😎🤯:

https://github.com/cartridge-gg/controller/assets/26515232/c8b20520-0453-47fa-8bd2-27f273fc1cc6

~~EDIT: just realized the weird bounce when opening the drawer. fixing it.~~ Fixed


tarrencev:
Looks slick! Didn't know you were hiding these skills

It would be really sick if when you scrolled up the pane header went all the way to the top of the frame before sticking

Also could be a cool behavior if you could scroll it open from the collapsed position, so you don't need to click the button

PR #369, opened 2024-06-14
Slot session creation page

kariy:
Screenshot 2024-06-18 at 3 09 57 PM

on success:

Screenshot 2024-06-18 at 3 10 39 PM

on failure:
Screenshot 2024-06-18 at 3 12 10 PM


broody:
Getting a few hydration errors on slot/session page

image

packages/keychain/src/pages/slot/session/index.tsx
```diff
@@ -0,0 +1,149 @@
...
+
+      fetch(url, {
+        body,
+        headers,
+        method: "POST",
+      })
+        .then((res) => {
+          res.status === 200
+            ? router.replace(`/slot/auth/success`)
+            : new Promise((_, reject) => reject(res));
```
broody, line 50: I think we need to `throw new Error()` here, otherwise this is uncaught

broody (commented):
Minor comment, otherwise works great!

packages/keychain/src/pages/slot/session/index.tsx
```diff
@@ -0,0 +1,156 @@
...
+      const url = decodeURIComponent(queries.callback_uri);
+      const body = JSON.stringify(session);
+      const headers = new Headers();
+      headers.append("Content-Type", "application/json");
+
+      fetch(url, {
+        body,
+        headers,
+        method: "POST",
+      })
```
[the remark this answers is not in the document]
broody: Would be good to fix this too

packages/keychain/src/pages/slot/session/index.tsx
kariy: i believe it should be caught with chain'd '.catch()`, no?

broody: Mmm if u try to make it fail, there's an uncaught error message. You could also do `return Promise.reject()`

kariy: ok i just have to make it a return statement, fixed it https://github.com/cartridge-gg/controller/pull/369/commits/bb3b07eee519d713d3260dc59baf6d29dc8a68e9 .

kariy:
> Getting a few hydration errors on slot/session page
> 
> image

fixed in 4a7f7f82cff017cd4074ccb0a7c5cc7d1a57ce1e

packages/keychain/src/pages/slot/session/index.tsx
```diff
@@ -0,0 +1,149 @@
...
```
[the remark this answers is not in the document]
kariy: @broody i think the callback url here will always be susceptible to user forgery (unless the full url is hardcoded, right it's not bcs ew rely on random port number). so i added a bouncer to validate the callback url here [6e98e30](https://github.com/cartridge-gg/controller/pull/369/commits/6e98e304e96287a0f6463e9ff0daf448ed8b0bda). Let me know if there's better alternative than this.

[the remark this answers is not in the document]
broody: Ah make sense!

PR #375, opened 2024-06-14
Off-chain signature

piniom:
Account_sdk functionality to support off chain signing of typed data. The signature can then be verified by locally hashing the message and calling the [is_valid_signature](https://github.com/cartridge-gg/controller/blob/54d14a52a3d65b4fec1238e5ff9ae4ed7dc18785/packages/contracts/controller/src/account.cairo#L202-L210) on the contract.

### DIFF
--- a/packages/account-wasm/src/lib.rs
+++ b/packages/account-wasm/src/lib.rs
@@ -5,7 +5,7 @@ use std::str::FromStr;
 
 use account_sdk::account::session::hash::{AllowedMethod, Session};
 use account_sdk::account::session::SessionAccount;
-use account_sdk::account::{AccountHashSigner, CartridgeGuardianAccount};
+use account_sdk::account::{AccountHashSigner, CartridgeGuardianAccount, MessageSignerAccount};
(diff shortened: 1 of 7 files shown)


packages/account-wasm/src/lib.rs
```diff
@@ -183,7 +183,11 @@ impl CartridgeAccount {
     }
 
     #[wasm_bindgen(js_name = signMessage)]
-    pub fn sign_message(&self) -> Vec<JsValue> {
-        unimplemented!("Sign Message not implemented");
+    pub async fn sign_message(&self, typed_data: String) -> Result<JsValue> {
```
tarrencev, line 186: should this be a string or a typed data structure?

broody: We could create a intermediate type TypedData for wasm but I feel like it's not necessary. Just need to go from the existing TypedData in JS to existing TypedData in rust, serde is probably the easiest

broody:
Tested on my end, works well! Not a big deal but I guess we don't support revision 0, will handle this in the frontend.

PR #433, opened 2024-07-02
breaking: bump `starknet-rs` to `0.11.0`

kariy:
Use `starknet::core::types::Felt` from now. 

This migration is needed to migrate Dojo to the latest `starknet-rs` version.

The contracts aren't recompiled to avoid possibly breaking stuff. Only the abis are regenerated using `cainome` with the new `Felt` type update.

**BREAKING**:

- bump `starknet-rs` to `0.11.0`
- migrate the old `FieldElement` to the new `Felt` type from `starknet-types-core`
- decided to not to just rename `Felt` -> `FieldElement` to avoid replacing every references of it so as to avoid confusion later

## Summary by CodeRabbit

- **Dependency Updates**
  - Updated several dependencies to their latest versions, including `starknet`, `starknet-crypto`, and `indexmap`.
  - Added new dependencies such as `num-traits`.

- **Code Improvements**
  - Replaced `FieldElement` with `Felt` across various implementations to align with the latest updates in the `starknet` ecosystem.
  - Updated method signatures and struct definitions to use `Felt`.
  - Modified method calls from `execute` to `execute_v1`.

- **CI/CD**
  - Updated Docker image version for the `rust-test` job in the GitHub Actions workflow.


kariy:
@piniom is it safe to recompile and regenerate abigens for the  contracts here ? actually i just need to regenerate abigen because need to migrate to the new `Felt` type, but i just did `make generate_artifacts` bcs im lazy.

piniom:
@kariy if you run `make`, then it will build the contract, generate the artifacts and generate abigen using cainome. It's safe i think.

kariy:
> @kariy if you run `make`, then it will build the contract, generate the artifacts and generate abigen using cainome. It's safe i think.

are the signatures created with the new build backwards compatible with the current build?

broody:
> are the signatures created with the new build backwards compatible with the current build?

Yep, we should be good. Compatibility breaks if / when we introduce this back tho https://github.com/cartridge-gg/controller/pull/426

broody (approved):

PR #434, opened 2024-07-02
Remove type module from account-wasm

JunichiSugiura:
Having `"type": "module"` option in `account-wasm` pkg causes ESM import error on Keychain (NextJS).


broody:
Ah nice, looks like it was merged in this PR https://github.com/cartridge-gg/controller/pull/436/files

PR #436, opened 2024-07-02
Add delegate account

notV4l:
Merge #419 & #420 

Add delegate account

related to https://github.com/cartridge-gg/internal/pull/1611


broody:
LGTM and is backwards compatible. @piniom we're trying to move fast so I'm gonna merge it, feel free to leave comments and we can address later

PR #444, opened 2024-07-05
feat(controller): add NCC for compiling controller into a standalone js

Larkooo:
for web pages without a bundler, like Unity games running on WebGL, we need to bundle packages into standalone source files, so this PR adds "ncc" and a compile directive to build the typescript files and their dependencies into a singular index.js file


Larkooo:
one thing to note though is that the last commit on @vercel/ncc was 5 months ago so it doesn't seem like it's being properly maintained. a great alternative to that would be rollup which is another bundler altogether

broody (approved):

PR #538, opened 2024-08-06
Configure @cartridge/ui-next in next example

JunichiSugiura:


JunichiSugiura:
hmm, seems ts-patch isn't working

PR #551, opened 2024-08-08
feat: added realmofra

okhaimie-dev:


broody:
@okhaimie-dev could you please also add ur id and primary color to the presets here:
https://github.com/cartridge-gg/controller/blob/main/packages/controller/src/presets.ts

okhaimie-dev:
@broody I just added the preset

PR #553, opened 2024-08-09
add openSetDelegate(address)

notV4l:


packages/connector/src/index.ts
```diff
@@ -68,6 +68,10 @@ class ControllerConnector extends Connector {
     return await this.controller.delegateAccount();
   }
 
+  async openSetDelegate(address: string) {
```
tarrencev, line 71: i think this could just be `setDelegate`, since if the session token is authorized, we could set it without the modal (in the future)

PR #556, opened 2024-08-11
Hook e2e test to CI

JunichiSugiura:


JunichiSugiura:
close in favor of #567

The reason: The example preview page fails to identify keychain preview URL when the branch name is too long.

PR #572, opened 2024-08-13
Test cross origin

broody:


examples/starknet-react-next/src/components/providers/StarknetProvider.tsx
```diff
@@ -51,16 +51,7 @@ const cartridge = new CartridgeConnector(
...
-    url:
-      !process.env.NEXT_PUBLIC_VERCEL_BRANCH_URL ||
-      process.env.NEXT_PUBLIC_VERCEL_BRANCH_URL.split(".")[0] ===
-        "cartridge-starknet-react-next"
-        ? process.env.XFRAME_URL
-        : "https://" +
-          (process.env.NEXT_PUBLIC_VERCEL_BRANCH_URL ?? "").replace(
-            "cartridge-starknet-react-next",
-            "keychain",
-          ),
```
JunichiSugiura, line 63: Oh this logic doesn't work? I think this code does not link to thge correct keychain preview link only when branch name is longer than certain length.

broody: Ah good to know, but yeah I tried it first without override and it wasn't working...

PR #592, opened 2024-08-16
Skip CreateSession if no policy is provided

JunichiSugiura:


broody:
Signup works great, we just need to do the same for login.

image

PR #598, opened 2024-08-20
Skip CreateSession if no policy is provided 

JunichiSugiura:
retry #592

### DIFF
--- a/packages/keychain/src/components/connect/Login.tsx
+++ b/packages/keychain/src/components/connect/Login.tsx
@@ -86,8 +86,8 @@ function Form({
           await doLogin(usernameField.value, credentialId);
           break;
         case LoginMode.Controller:
-          if (policies.length === 0) {
-            throw new Error("Policies required for controller ");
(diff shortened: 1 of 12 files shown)


packages/keychain/src/components/connect/Login.tsx
```diff
@@ -81,17 +81,10 @@ function Form({
...
-          if (policies.length === 0) {
-            throw new Error("Policies required for controller ");
-          }
-
-          await controller.approve(origin, expiresAt, policies);
-          break;
+      if (mode === LoginMode.Controller && policies.length > 0) {
+        await controller.approve(origin, expiresAt, policies);
+      } else {
+        await doLogin(usernameField.value, credentialId);
```
broody, line 87: @JunichiSugiura I made small change here to make login require webauthn authentication if no policies. Otherwise ppl can log into others account, any txn will fail but just feels weird.

broody (approved):

PR #635, opened 2024-08-29
Show verified badge

### LINKED ISSUE #594: Add session key verification badge to the controller
Lets manage and surface a verified badge for session keys in the controller. Verified sessions/games should be maintained in a public github repo. people will send a PR to add their game to the verified list. Clicking on the verified badge should take you to the repo. 

Here are the designs

--- 

The verified badge appears in two places

On the log in page where it expands with the session tray

image
image

And on the dedicated session page (ive updated the design here to give the session details a little more room) 

image

---

[Figma Link](https://www.figma.com/design/OSbzNkyR9UYV4Erz9oF9JJ/Controller?node-id=3692-35410&t=MvzrNy0BFfo3gzch-1)

JunichiSugiura:
close #594

It might make sense to move all preset theme to`verified.ts` once we list all url.


packages/controller/src/verified.ts
```diff
@@ -0,0 +1,7 @@
+export const verified: { id: string; url: string }[] = [
+  { id: "flippyflop", url: "https://flippyflop.vercel.app" },
```
tarrencev, line 2:
```suggestion
  { id: "flippyflop", url: "https://flippyflop.gg" },
```

PR #647, opened 2024-09-01
Fix contract not found on local dev

broody:


packages/keychain/src/utils/connection/execute.ts
```diff
@@ -78,7 +78,8 @@ export function executeFactory({
           } catch (error) {
             if (
               error instanceof Error &&
-              error.message.includes("ContractNotFound")
+              (error.message.includes("ContractNotFound") ||
+                error.message.includes("is not deployed"))
```
tarrencev, line 82: we should make this more specific to the account address since we don't want to trigger the deploy flow for a deployed account calling an underpowered contract

PR #650, opened 2024-09-03
Build wasm at runtime

tarrencev:
Looks like we will need to move to this to enable this PR:

https://vercel.com/guides/how-can-i-use-github-actions-with-vercel


steebchen:
we can close this ig @tarrencev?

PR #664, opened 2024-09-05
Add profile iframe

JunichiSugiura:
Add `Open Profile` button to Controller example.

### One question
- https://github.com/cartridge-gg/controller/pull/664#pullrequestreview-2313514315


JunichiSugiura:
Auto connect does not work anymore since keychain connection takes longer than used to be for some reason.
Seems like a race condition to me.

Screenshot 2024-09-18 at 13 19 32

packages/controller/src/index.ts
```diff
@@ -17,39 +13,52 @@ import {
...
+import { RPC_SEPOLIA } from "./constants";
 
 class Controller {
-  private url: URL;
   private policies: Policy[];
   private paymaster?: PaymasterOptions;
-  private connection?: Connection<Keychain>;
-  private modal?: Modal;
   public keychain?: AsyncMethodReturns<Keychain>;
+  public profile?: AsyncMethodReturns<Profile>;
```
JunichiSugiura, line 30: I wonder if `controller.keychain` has to be public 🤔

JunichiSugiura, line 30: I wonder if `controller.keychain` has to be public 🤔

broody: Yeah good point, I think both keychain and profile should be private

broody (approved):
This make it so much cleaner! Don't see anything that would cause backward incompatibility so LGTM

packages/controller/src/index.ts
```diff
@@ -17,39 +13,52 @@ import {
...
+  private iframes: IFrames;
   public rpc: URL;
   public account?: AccountInterface;
 
   constructor({
     policies,
     url,
+    profileUrl,
```
broody, line 38: Mmm do we need to make this user configurable? Could it just be const?

JunichiSugiura: Good point. Yes, I agree that there's no need for game dev to set this url. But we want it as a Controller dev to switch between localhost and production API I think. Maybe we can introduce env to Controller but I'm not keen to do so since it's a library not app.

PR #689, opened 2024-09-11
Setup Storybook for Keychain

JunichiSugiura:
### Note
I had to extract `useConnectionValue` from provider module since importing `@cartridge/penpal` in storybook causes error.

Screenshot 2024-09-11 at 15 57 11


JunichiSugiura:
For some reason, Storybook does not render decorator when `<ConnectionProvider />` is rendered 🫠
Weird because other providers such as `<ChakraProvider />` or `<ControllerThemeProvider />` works fine...

tarrencev (approved):

PR #691, opened 2024-09-11
Fetch balance on init

broody:


JunichiSugiura (approved):
Ohhh, thanks for the catch!

PR #716, opened 2024-09-13
Show decoded chainId with Slot icon

JunichiSugiura:
[Storybook - `<NetworkStatus />`](https://keychain-storybook-git-featkeychain-network-name.preview.cartridge.gg/?path=/docs/components-layout-container-header-networkstatus--docs)


packages/keychain/src/components/layout/Container/Header/NetworkStatus.tsx
```diff
@@ -0,0 +1,51 @@
...
+    <Button
+      size="sm"
+      bg="solid.primary"
+      fontSize="xs"
+      fontFamily="Inter"
+      _hover={{ bg: "solid.primary", opacity: 0.7 }}
+      iconSpacing={1.5}
+      leftIcon={
+        <Circle size={5} bg="solid.primary">
+          {(() => {
```
broody, line 27: Controller on localhost coming soon. I think we can first check rpcUrl and see if localhost then display Katana? Katana by default spins up with `SN_SEPOLIA`

broody (approved):

packages/keychain/src/components/layout/Container/Header/NetworkStatus.tsx
JunichiSugiura: Oh good point, totally ! I'll add that

PR #717, opened 2024-09-13
Update build process, adapt wasm and keychain CI

steebchen:


.github/workflows/e2e.yml
```diff
@@ -1,27 +1,28 @@
...
-#           retention-days: 30
+ name: End-to-End tests
+ on:
+   push:
+     branches: [main]
+   pull_request:
+     branches: [main]
+ jobs:
+   test:
+     if: always() == 'skip' # temporarily skip cc @jun
```
JunichiSugiura, line 9: Yes 🙏

turbo.json
```diff
@@ -2,6 +2,9 @@
   "$schema": "https://turbo.build/schema.json",
   "ui": "tui",
   "tasks": {
+    "build-wasm": {
+      "outputs": ["packages/account-wasm/pkg/**"]
+    },
```
JunichiSugiura, line 7: It's prob simpler to rename the task to build:deps in `packages/account-wasm/package.json` so `pnpm turbo build:deps` handles the all lib build

JunichiSugiura (changes requested):

turbo.json
steebchen: we probably don't want to build wasm everywhere though as it has more deps and requires more resources

JunichiSugiura: ah I see, I guess that's why `pkg` is committed. Makes sense

.tool-versions
```diff
@@ -1,3 +1,4 @@
 scarb 2.7.0
 starknet-foundry 0.18.0
 nodejs v20.11.1
+pnpm 9.7.0
```
JunichiSugiura, line 4: pnpm version should be managed by corepack

```sh
# once Node.js is installed via asdf or whatever vm,
corepack enable pnpm
```

JunichiSugiura (changes requested):
Just one thing about pnpm vm, otherwise it's working fine locally!

.tool-versions
steebchen: fair

steebchen: can we have it here anyway? I am using asdf for pnpm

JunichiSugiura: I prefer to avoid duplication of versioning

https://github.com/cartridge-gg/controller/blob/1bc2065f4cad65f23211959557e618f2215109b6/package.json#L3

JunichiSugiura (approved):

PR #719, opened 2024-09-13
Prompt session creation on execute if mismatch

broody:
Not implemented here is actually revoking previous session

image

### DIFF
--- a/packages/keychain/src/components/ConfirmTransaction.tsx
+++ b/packages/keychain/src/components/ConfirmTransaction.tsx
@@ -1,15 +1,17 @@
-import { useMemo } from "react";
+import { useMemo, useState } from "react";
 import { Policy, ResponseCodes } from "@cartridge/controller";
 import { Content, FOOTER_MIN_HEIGHT } from "components/layout";
 import { TransactionDuoIcon } from "@cartridge/ui";
 import { useConnection } from "hooks/connection";
 import { Policies } from "components/Policies";
 import { ExecuteCtx } from "utils/connection";
-import { num } from "starknet";
+import { addAddressPadding, num } from "starknet";
 import { ExecutionContainer } from "components/ExecutionContainer";
+import { CreateSession } from "./connect";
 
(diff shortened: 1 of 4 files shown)


packages/keychain/src/components/ConfirmTransaction.tsx
```diff
@@ -23,14 +25,40 @@ export function ConfirmTransaction() {
...
+      })),
     [ctx.transactions],
   );
+  const updateSession = useMemo(() => {
+    if (policiesUpdated) return false;
+
+    const txnsApproved = transactions.every((transaction) =>
+      policies.some(
+        (policy) =>
+          addAddressPadding(policy.target) ===
```
JunichiSugiura, line 45: oh i didn't know about this util func. i should use that for formatAddress func.

packages/keychain/src/components/ConfirmTransaction.tsx
```diff
@@ -23,14 +25,40 @@ export function ConfirmTransaction() {
...
+      ),
+    );
+
+    return !account.hasSession(ctx.transactions) && txnsApproved;
+  }, [transactions, policiesUpdated, policies, account, ctx.transactions]);
+
+  if (updateSession) {
+    return (
+      <CreateSession
+        isUpdate={true}
```
JunichiSugiura, line 57: nit: you can omit `={true}`

JunichiSugiura (approved):
LGTM

JunichiSugiura:
is there a case when game dev want to trigger just one tx with specific method(i.e. transfer) which doesn't want to include in session?

broody:
> is there a case when game dev want to trigger just one tx with specific method(i.e. transfer) which doesn't want to include in session?

Yup, if the call is not in requested policies then we don't display Create Session and fallback to weabuthn execution.

https://github.com/cartridge-gg/controller/pull/719/files#diff-b27c08a79b37c2b11dfb61c56eee2b8e19659f140a75fa3c43d48f2d88c6f558R51

Issue #671, opened 2024-09-07
Update the error state for "username already exists"


clicksave:
Lets make it easier to move from log in to sign up when a user inputs a invalid/taken username

<img width="1168" alt="image" src="https://github.com/user-attachments/assets/3530c69f-4b53-4032-ae37-3c214134e04e">

Doing a few things here 

- [ ] removes the alert icon
- [ ] changes the copy
- [ ] adds an inline 'Log in button' 

<img width="858" alt="image" src="https://github.com/user-attachments/assets/da520c70-bcb0-48be-a679-01155c7f0509">

[Figma](https://www.figma.com/design/OSbzNkyR9UYV4Erz9oF9JJ/Controller?node-id=3794-9077&t=BNkHGn3BbXI9OmBG-1)

---

Lets use a similar pattern on sign up as well 

<img width="1120" alt="image" src="https://github.com/user-attachments/assets/46f32fce-af4b-4f6f-aaa5-ab27a020cec7">

[Figma](https://www.figma.com/design/OSbzNkyR9UYV4Erz9oF9JJ/Controller?node-id=5987-50579&t=BNkHGn3BbXI9OmBG-1)

bitfalt:
hii! I'd love to work on this. Just to get the task clear, I would need to update the error state and make it like the Figma designs which would be to fulfill the tasks you mentioned:

- [ ] Remove the alert icon
- [ ] Changes the copy (I would like to know what you mean exactly by this)
- [ ] Add an inline 'Log in' button.

clicksave:
gm @bitfalt you can have this ticket if you'd like but I just wanted to confirm. The controller isn't an open source project so theres no bounty for this work. lmk if you still want it

bitfalt:
okay okay, no worries! I'll still be happy to help :D

clicksave:
Awesome! ok, to answer your question, I'm referring to the contents of the error message 

- [ ] The copy for the error state on the Sign Up page currently reads "Controller with username not found" it should instead read "Username not found" 
- [ ] The copy for the error state on the Log In page currently reads "Account already exists" it should instead read "Sorry, [username] already exists." 

You can test it yourself here if you'd like 
https://flippyflop.gg/

clicksave:
> hii! I'd love to work on this. Just to get the task clear, I would need to update the error state and make it like the Figma designs which would be to fulfill the tasks you mentioned:
> 
> * [ ]  Remove the alert icon
[...]

PR #723, opened 2024-09-14
Update the error state for "username already exists"

bitfalt:
Solves #671 

- [x] Change the error message in Signup and Login
- [x] Add the button to Signup modal
- [ ] Add the button to Login modal

### DIFF
--- a/packages/keychain/src/components/connect/Signup.tsx
+++ b/packages/keychain/src/components/connect/Signup.tsx
@@ -210,6 +210,19 @@ export function Signup({
               setUsernameField((u) => ({ ...u, value: "" }));
             }}
           />
+          {usernameField.error && usernameField.error.includes(
+            `Sorry, ${usernameField.value} already exists.`
+          ) && (
+            <div>
+              {usernameField.error}
+              <RegistrationLink
+                description="Is this you?"
+                onClick={() => onLogin(usernameField.value)}
+              >
+                Sign Up
+              </RegistrationLink>
+            </div>
+          )}
         </Content>
 
         <Footer isSlot={isSlot} isSignup showCatridgeLogo>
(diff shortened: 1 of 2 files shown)


bitfalt:
Hi! I'm trying to test the button I added to check if it displays correctly and then start modifying the styling according to the Figma. However, I'm getting some problems when trying to open the modal in `localhost:3002`.

I'm getting the following error:
image

From the error I see that it is trying to make a request to `localhost:8001/x/starknet/sepolia?paymaster=true`. However I don't have anything running in `localhost:8001`. I assume I need to run the local devnet, but I was just wondering if there is a specific configuration or something needed.

I followed all the instructions in the README regarding Frontend Development.

clicksave:
cc @broody any guidance on the issue Daniel is running into?

broody:
@bitfalt thanks for the PR! Sorry, but the way we're setup currently is this requires some of our backend services to be running locally. I don't think there is a good alternative rn.

Let me help u push this PR across tho, again thanks for trying to tackle this!

bitfalt:
okay okay! Could I dm you on Telegram or Discord to have better communication? I might need some guidance regarding proper implementation. I already modified the messages, but I usually struggle with setting up frontend related stuff, so some proper guidance would be amazing! :D

broody:
> okay okay! Could I dm you on Telegram or Discord to have better communication? I might need some guidance regarding proper implementation. I already modified the messages, but I usually struggle with setting up frontend related stuff, so some proper guidance would be amazing! :D

Yeah absolutely, drop by discord here https://discord.gg/cartridge

Or you can do me on telegram at @broody_eth

tarrencev:
@bitfalt thanks for this contribution. unfortunately our plans have changed here and we have an alternate path we're working towards

PR #725, opened 2024-09-16
Fix register session estimate fee calldata

broody:
Fixes this error when policies >2
cc @kariy 
image


kariy (approved):
it works! thanks.

for some reason now i got a insufficient fee error 

```
Transaction reverted. reason=Insufficient max fee: max fee: 277300000000000, actual fee: 417400000000000.
```

but its unrelated to the PR, so should be good to go

PR #733, opened 2024-09-17
Only build related packages

JunichiSugiura:
Make it easier to identify where the build error comes from.

i.e. keychain deploy job should not fail when there's a build error on keychain-storybook
Screenshot 2024-09-17 at 12 04 39


JunichiSugiura:
btw, i'm wondering if this build step is necessary since these build scripts are stated on Vercel side which I assume would be pulled by `pnpm vercel pull` command in following step. @steebchen

JunichiSugiura:
close in favor of #731

PR #738, opened 2024-09-17
fix: fee estimation multiplier

kariy:
remove unused fee multiplier field in account-wasm/account-sdk. is it unused because setting the `fee_estimate_multiplier` here is pretty much a no-op because we call the `estimate_fee()` and not `send()`.

https://github.com/cartridge-gg/controller/blob/c283c0e3671b732b54ee8eea11b4da970d767efd/packages/account_sdk/src/controller.rs#L263-L268

this PR removes all references of fee multiplier on account-wasm side and compute it on javascript end instead.

### DIFF
--- a/packages/account-wasm/src/lib.rs
+++ b/packages/account-wasm/src/lib.rs
@@ -177,7 +177,6 @@ impl CartridgeAccount {
     pub async fn estimate_invoke_fee(
         &self,
         calls: Vec<JsCall>,
-        fee_multiplier: Option<f64>,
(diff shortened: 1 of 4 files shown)


broody (approved):
Ah we do account for Webauthn gas now, but I guess the 1.5x wasn't taking affect previously. nice catch!

https://github.com/cartridge-gg/controller/blob/c283c0e3671b732b54ee8eea11b4da970d767efd/packages/account_sdk/src/controller.rs#L275

PR #753, opened 2024-09-20
fix: use raw string to avoid windows path considered as escaping

glihm:
Related to dojo error: https://github.com/dojoengine/dojo/actions/runs/10948582548/job/30400215646?pr=2454.


glihm:
Tackled in #757.

PR #784, opened 2024-09-26
Fix copying formatted address

broody:
Copy address was copying `0x1234....4568`


JunichiSugiura (approved):
oh sorry, it's a bug on formatAddress then

PR #794, opened 2024-10-01
Add GraphQL configuration and schema pull script

steebchen:
Created a new .graphqlrc.yml to define schema and document paths. Added a pull_schema.sh script to automate the GraphQL schema introspection and saving it locally. This setup will streamline the development process involving GraphQL queries.

### DIFF
--- a/.graphqlrc.yml
+++ b/.graphqlrc.yml
@@ -1,0 +1,2 @@
+schema: schema.json
+documents: packages/**/*.graphql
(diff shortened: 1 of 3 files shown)


broody (approved):
LGTM, tho keychain currently uses graphql-codegen. I guess the idea is eventually we migrate gql queries that to account_sdk as well and leverage this schema.json

PR #824, opened 2024-10-03
Unmocking guardian

piniom:


packages/account_sdk/src/controller.rs
```diff
@@ -65,6 +66,23 @@ where
...
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    pub fn with_guardian(
+        app_id: String,
+        username: String,
+        class_hash: Felt,
+        provider: P,
+        owner: Signer,
+        guardian: Option<Signer>,
```
tarrencev, line 82: im not sure if we want this here. since the sdk would never be it's own guardian. the guardian signer would come from the cartridge backend

packages/contracts/controller/src/account.cairo
```diff
@@ -454,10 +489,20 @@ mod CartridgeAccount {
         fn assert_valid_span_signature(
             self: @ContractState, hash: felt252, signer_signatures: Array<SignerSignature>
         ) {
-            assert(signer_signatures.len() <= 2, 'invalid-signature-length');
+            assert(signer_signatures.len() >= 1, 'empty-signatures');
```
tarrencev, line 492:
```suggestion
            assert(signer_signatures.len() <= 1, 'empty-signatures');
```

PR #826, opened 2024-10-03
Fix register session redirecting immediately

broody:


packages/keychain/src/pages/session.tsx
```diff
@@ -151,11 +168,12 @@ export default function Session() {
 function sanitizeCallbackUrl(url: string): URL | undefined {
   try {
     const parsed = new URL(url);
+    const allowedHostnames = ["localhost"];
```
tarrencev, line 171: Hmm will this prevent callbacks to i.e. the telegram app?

Larkooo: This is only for callbacks so it wouldn't apply. Telegram uses redirects.

tarrencev (approved):

PR #859, opened 2024-10-14
Fix self deployment

broody:


packages/account-wasm/src/account.rs
```diff
@@ -52,7 +52,7 @@ impl CartridgeAccount {
         let controller = Controller::new(
             app_id,
             username.clone(),
-            CONTROLLERS[&Version::V1_0_4].hash,
+            CONTROLLERS[&Version::LATEST].hash,
```
tarrencev, line 55: hmm i don't think we want to default to latest here. better to be explicit

PR #867, opened 2024-10-16
Feat/guardian-unmock-1: Contracts

piniom:
The first of a couple of stacked PRs on the guardian implementation.
Added guardian mechanism to the contract code.

- A valid guardian signature is now required when a guardian is set on the account
- When the guardian is `None` a valid guardian signature is allowed but not required
- When registering (or caching) a session, a guardian is taken into account
- When registering a session, a user provides an owner which is associated with the particular session (since there might be many), but the guardian is fetched from the account storage.

## Summary by CodeRabbit

- **New Features**
	- Introduced a method to check if a session is registered, enhancing session management capabilities.
	- Updated the session structure to include a new guardian key identifier for improved data handling.

- **Bug Fixes**
	- Improved serialization and deserialization processes for the updated session structure.


packages/contracts/controller/src/account.cairo
```diff
@@ -147,6 +152,7 @@ mod CartridgeAccount {
 
     #[storage]
     struct Storage {
+        guardian_guid: Option<felt252>,
```
tarrencev, line 155: wondering if this could be `guardian_guid: felt252` and the 0 value could be treated none?

packages/contracts/controller/src/account.cairo
```diff
@@ -425,8 +449,14 @@ mod CartridgeAccount {
...
         ) -> bool {
-            assert(signer_signatures.len() <= 2, 'invalid-signature-length');
-            self.is_valid_owner_signature(hash, *signer_signatures.at(0))
+            if signer_signatures.is_empty() {
+                return false;
+            }
+            if signer_signatures.len() > 2 {
+                return false;
+            }
+            self.is_valid_owner_signature(hash, *signer_signatures.at(0)) &&
```
tarrencev, line 458: can we do `assert(signer_signatures.len() == 2, ...)` and have the dummy guardian sig that isn't validated if necessary?

packages/contracts/controller/src/account.cairo
```diff
@@ -454,10 +504,21 @@ mod CartridgeAccount {
         fn assert_valid_span_signature(
             self: @ContractState, hash: felt252, signer_signatures: Array<SignerSignature>
         ) {
-            assert(signer_signatures.len() <= 2, 'invalid-signature-length');
+            assert(signer_signatures.len() >= 1, 'empty-signatures');
```
tarrencev, line 507:
```suggestion
            assert(signer_signatures.len() != 0, 'empty-signatures');
```

but i think we can do `== 2` if we always send the dummy sig

packages/contracts/controller/src/account.cairo
```diff
@@ -454,10 +504,21 @@ mod CartridgeAccount {
...
             assert(
                 self.is_valid_owner_signature(hash, *signer_signatures.at(0)), 'invalid-owner-sig'
             );
+            match self.guardian_guid.read() {
+                Option::Some(_) => {
+                    assert(signer_signatures.len() >= 2, 'no-guardian-sig');
```
tarrencev, line 513: when are there more than 2 sigs?

packages/contracts/controller/src/account.cairo
piniom: It could, less readable, but perhaps more efficient

packages/contracts/controller/src/account.cairo
piniom: why would we require a guardian signature when there is no guardian on the account? It think this would actually hit performance more than the optionable storage, since we are deserializing the signatures.

packages/contracts/controller/src/account.cairo
piniom: it's a sanity check. i think it's good it's here. we want to make sure the user doesn't pass more than two signatures since there is no reason to do so. 
Argent's imlementation also checks there are no redundant signatures
https://github.com/argentlabs/argent-contracts-starknet/blob/1352198956f36fb35fa544c4e46a3507a3ec20e3/src/presets/argent_account.cairo#L963

tarrencev: in this case assert will pass if there are 3 signatures though

packages/contracts/controller/src/account.cairo
tarrencev: fair enough

packages/contracts/controller/src/account.cairo
tarrencev: imo it would be a bit more readable since we do `unwrap_or(0);` so many places and essentially treat is as that

packages/contracts/controller/src/account.cairo
piniom: Line 514 asserts that, i wanted to provide separate error messages in each case

packages/contracts/controller/src/account.cairo
piniom: marking as resolved since we decided against that

packages/contracts/controller/src/account.cairo
piniom: I partially agree :)). All the unwraps happen in the session component, so the main account is more readable with an option. We could cache the sessions using (felt, Option<felt>, felt) instead of (felt, felt, felt), but we break compatibility with the sessions stored pre-guardian. 
wdyt? I'm not advocating for any of the two to be clear (felt everywhere, or optionable storage)

tarrencev: i think lets roll with felt everywhere, it also avoids a storage slot

piniom: ok!

PR #868, opened 2024-10-16
Feat/guardian-unmock-2: `TransactionWaiter`

piniom:
The second of a couple of stacked PRs on the guardian implementation.
Fixed a bug where a `TransactionWaiter` would return `Ok(...)` even thou a transaction was reverted and `must_succeed` was set to `true`.


piniom:
Duplicate of https://github.com/cartridge-gg/controller/pull/871

PR #869, opened 2024-10-16
Feat/guardian-unmock-2: Proxy and `account_sdk` tests

piniom:
The second of a couple of stacked PRs on the guardian implementation.
Changed the cartridge proxy to inject a guardian signature:
- when authorizing with a regular key
- when authorizing with a session
- when estimating the fee


tarrencev (approved):
nit on naming otherwise lgtm

PR #884, opened 2024-10-18
Session off chain signature account_sdk changes

piniom:


packages/account-wasm/src/types/policy.rs
```diff
@@ -25,19 +25,21 @@ impl TryFrom<JsValue> for Policy {
...
     type Error = EncodingError;
 
-    fn try_from(value: Policy) -> Result<Self, Self::Error> {
-        Ok(Self {
-            contract_address: Felt::from_str(&value.target)?,
-            selector: get_selector_from_name(&value.method).unwrap(),
-        })
+    fn try_from(_value: Policy) -> Result<Self, Self::Error> {
+        todo!();
+        // Ok(Self {
```
piniom, line 30: Should I modify this accordingly, or will the other guys do it to their liking?

tarrencev: Yes let's update these types

packages/account_sdk/src/account/session/raw_session.rs
```diff
@@ -65,63 +61,4 @@ impl SessionHash for RawSession {
...
-            session,
-            cache_authorization,
-            session_authorization,
-            session_signature,
-            guardian_signature,
-            proofs,
-        })
-    }
-}
+pub type RawSessionToken = crate::abigen::controller::SessionToken;
```
tarrencev, line 64: nice

Issue #825, opened 2024-10-03
Players need a way to manage their assets


clicksave:
This a a big ticket. 

We need to add a way for users to manage (i.e. send) the assets in their controller. 

--- 

Starts with the profile page where users can see their assets
<img width="1744" alt="image" src="https://github.com/user-attachments/assets/71c0579b-35f0-42f9-9ee6-34e0a3f3502f">

Assets are broken down into two categories 

Tokens and NFTS

---

Tokens include 

Credits, ETH, Whitelisted ERC20s

Clicking on one of these items takes you to a detail page where you are able to view the token balance

<img width="1574" alt="image" src="https://github.com/user-attachments/assets/6524c3f2-eb9d-4c41-9552-80a10091a776">

---

Clicking send takes you to a form where you specify the destination address and amount you want to send

<img width="1190" alt="image" src="https://github.com/user-attachments/assets/66b01027-6611-4452-8143-ca058c24ee54">

---

Finally, after inputing the amount and destination, you review and confirm the transaction

<img width="1793" alt="image" src="https://github.com/user-attachments/assets/bcf9ae15-1a22-44cb-8544-c50f8bee53d2">

---

Sending NFTs is a little different

Players click a collection and are taken to a collection page. There they can select an NFT

<img width="1564" alt="image" src="https://github.com/user-attachments/assets/8de7d211-ed37-449c-a273-5a6d6935e192">

They are taken to the asset preview page

<img width="1221" alt="image" src="https://github.com/user-attachments/assets/a396e9fd-10ef-4b5a-8777-0c6815d286ad">

Clicking 'Send' here prompts the user to specify a destination address 

<img width="1128" alt="image" src="https://github.com/user-attachments/assets/a311d569-193f-4c95-8f47-5ef15152c443">

Finally, they review the transaction and submit it

<img width="1697" alt="image" src="https://github.com/user-attachments/assets/ba976c73-3ee2-4ad7-a6b1-a867e511ec9f">

---

You can also bulk send NFTS (lower priority than the rest of this) so maybe worth disregarding for now but heres how it works. 

https://www.figma.com/design/OSbzNkyR9UYV4Erz9oF9JJ/Controller?node-id=6660-7453&t=NmIDDW3Ht7LJC0es-1

PR #894, opened 2024-10-21
ERC20 token details page

JunichiSugiura:
#825 

- Add details page for ERC20 tokens
- Move GraphQL related logic from `@cartridge/keychain` to `@cartridge/utils` to share with profile
  - Cleanup `@graphql-codegen/*` deps from keychain
- `NEXT_PUBLIC_API_URL` -> `NEXT_PUBLIC_CARTRIDGE_API_URL`
- Pass GraphQL endpoints via context instead of directly reference env in custom fetcher
- Remove `generated.ts` files from git
  - `pnpm utils build:deps` generates them locally


broody (approved):
Make sense! not sure if these generated gql has the credits stuff, maybe need to regen with latest

PR #895, opened 2024-10-21
📜 Feat(quest)/integration

bal7hazar:
Integration of the trophies in the profile page.

Some info:
- `Quest` has been renamed into `Trophies` because we actually focus on achievements and not on the flow to unlock them (only the description is here to help the user, but the quest itself will be handled on the game side).
- The data is hardcoded for now, the next PR regarding trophies will be the connection with the data.
- I have added a dedicated graphql query for the AccountName, but actually the field `name` is empty and we use the field `id`, but in the future we should rely on the name and not the id.
- I have modified the default behavior of the ScrollArea primitive to have the scrollbar hidden always except on scroll (and fad out after few hundred ms).
- I have added few icons coming from figma
- I have added few colours without refactoring it all, but at some point we should consider to realign it with figma and remove the offset (secondary -> primary, quaternary -> secondary, etc.)


packages/profile/src/components/navigation.tsx
```diff
@@ -31,7 +32,7 @@ function Item({
   const location = useLocation();
 
   const isActive =
-    location.pathname == `/${variant}` ||
+    location.pathname.includes(`/${variant}`) ||
```
JunichiSugiura, line 35: nit: `.startsWith()`

packages/profile/src/components/trophies/index.tsx
```diff
@@ -0,0 +1,190 @@
...
+    { address: address || self },
+    {
+      enabled: false,
+      onSuccess: async (data: AccountNameQuery) => {
+        setUsername(data.accounts?.edges?.[0]?.node?.id ?? "Anonymous");
+      },
+    },
+  );
+
+  const fetchName = useCallback(async () => {
```
JunichiSugiura, line 112: nit: is this necessary? maybe destructure as `fetchName` on line 102?

packages/profile/src/components/trophies/leaderboard.tsx
```diff
@@ -0,0 +1,84 @@
...
+  players,
+  address,
+}: {
+  players: Player[];
+  address: string;
+}) {
+  return (
+    <div className="flex flex-col gap-y-px rounded-md overflow-hidden">
+      {players
+        .sort((a, b) => a.rank - b.rank)
```
JunichiSugiura, line 20: nit: I assume the sort exists for mocked data? graphql response should be already sorted or has sort param.

packages/profile/src/components/trophies/leaderboard.tsx
```diff
@@ -0,0 +1,84 @@
...
+  Icon,
+}: {
+  username: string;
+  self: boolean;
+  Icon: React.ComponentType<StateIconProps> | undefined;
+}) {
+  return (
+    <div className="flex items-center gap-x-2">
+      {!!Icon && <Icon size="default" variant="line" />}
+      {!Icon && <SpaceInvaderIcon size="default" variant="line" />}
```
JunichiSugiura, line 71: nit: ternary operator might be cleaner.

packages/profile/src/components/trophies/pinned.tsx
```diff
@@ -0,0 +1,65 @@
...
+      </CardHeader>
+    </Card>
+  );
+}
+
+function Title({ title, empty }: { title: string; empty?: boolean }) {
+  const overflow = useMemo(() => title.length > 27, [title]);
+  const content = useMemo(() => {
+    if (!overflow) return title;
+    return title.slice(0, 24) + "...";
```
JunichiSugiura, line 49: Can it be handled by `text-overflow` css?

https://tailwindcss.com/docs/text-overflow#truncate

packages/ui-next/src/themes/dark.css
```diff
@@ -1,25 +1,31 @@
 @layer base {
   .dark {
-    --background: 135 8% 9%; /* bg */
+    --background: 135 15% 10%; /* bg.background */
```
JunichiSugiura, line 3: 🙏🙏🙏

JunichiSugiura (approved):
It was fast and looks clean, LGTM !

packages/profile/src/components/trophies/pinned.tsx
bal7hazar: I was thinking it was not possible because I wanted the ellipsis on 2nd line, but actually I was able to make it works thanks to `line-clamp` that I didnt know, good catch

```js
break-words text-ellipsis line-clamp-2
```

Issue #829, opened 2024-10-04
Collapse viewport height when keyboard expands


clicksave:
This applies to the log in and signup pages and fixes an issue where this breaks on certain browsers/devices

see -> 
<img width="938" alt="image" src="https://github.com/user-attachments/assets/0d232c38-a825-408d-81b2-afbf02b92fd6">

---

<img width="781" alt="image" src="https://github.com/user-attachments/assets/0f34ffcf-d9a8-4b8f-8b80-062ce54774a7">

---

https://www.figma.com/design/OSbzNkyR9UYV4Erz9oF9JJ/Controller?node-id=6665-12733&t=b8uRkMXcOpYzcxl6-1

PR #902, opened 2024-10-23
Fix viewport height too small

broody:
Fixes https://github.com/cartridge-gg/controller/issues/829

We basically change layout variant when height is too small.

@JunichiSugiura I reduced layout variants from `connect, error, settings, default` to just `expanded, compressed`, those were the two main ones I see, lemme know if something's missing

### DIFF
--- a/packages/keychain/src/components/layout/Container/index.tsx
+++ b/packages/keychain/src/components/layout/Container/index.tsx
@@ -46,7 +46,7 @@ export const FOOTER_HEIGHT = 40;
 export const PORTAL_WINDOW_HEIGHT = 600;
 
 function Wrapper({
-  variant = "full",
+  variant = "expanded",
   children,
   ...rest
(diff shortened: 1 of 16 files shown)


packages/keychain/src/components/layout/Container/index.tsx
```diff
@@ -122,12 +122,12 @@ type LayoutContextValue = {
   };
 };
 
-type LayoutVariant = "default" | "connect" | "settings" | "error";
+export type LayoutVariant = "expanded" | "compressed";
```
JunichiSugiura, line 125: 👍

JunichiSugiura (approved):
LGTM !

JunichiSugiura:
ah forgot to mention, the original issue is also to move the footer above the keyboard on mobile. But it requires to use [Visual Viewport API](https://developer.mozilla.org/en-US/docs/Web/API/Visual_Viewport_API) i believe. Not straight forward

PR #940, opened 2024-10-25
Fix connect interface

broody:


packages/keychain/src/utils/connection/connect.ts
```diff
@@ -11,7 +11,7 @@ export function connectFactory({
   setContext: (context: ConnectionCtx) => void;
 }) {
   return (origin: string) =>
-    (rpcUrl: string): Promise<ConnectReply> => {
+    (_: Policy[], rpcUrl: string): Promise<ConnectReply> => {
```
tarrencev, line 14: I think we want to add back the `setPolicy` stuff here i order to make it backwards compatible right?

broody: Ah I had assumed u were pulling those from url, yeah let me add them back

tarrencev: We are and I think contorller 0.3.46 actually adds them to the url already too right?

Maybe we don't need it then

PR #954, opened 2024-10-29
Fix erc20 balance fetching wrong address

broody:


JunichiSugiura (approved):
oh thank you for the catch 🙏

PR #956, opened 2024-10-30
build connector & controller as esm using tsup

notV4l:


packages/controller/src/session/provider.ts
```diff
@@ -29,7 +29,7 @@ export default class SessionProvider extends BaseProvider {
   protected _username?: string;
   protected _redirectUrl: string;
   protected _policies: Policy[];
-  public account?: SessionAccount;
+  // public account?: SessionAccount;
```
tarrencev, line 32:
```suggestion
```

lets remove if it is unused

broody (approved):

PR #976, opened 2024-11-04
Fix search param handling

JunichiSugiura:
- Keep initial search params passed from iframe in connection context
  - prevent loosing them in page navigation
  - fixes https://github.com/cartridge-gg/controller/pull/970#discussion_r1825330551
- Add `slot` option to provide Slot project name
- Keep namespace option for Trophy page
- Hide activities for now

### DIFF
--- a/packages/profile/src/hooks/account.ts
+++ b/packages/profile/src/hooks/account.ts
@@ -18,7 +18,11 @@ export function useUsername({ address }: { address: string }) {
 }
 
 export function useAccount() {
-  const params = useParams<{ username: string; namespace?: string }>();
+  const params = useParams<{
+    username: string;
+    project?: string;
+    namespace?: string;
+  }>();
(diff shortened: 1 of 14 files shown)


packages/profile/src/hooks/account.ts
```diff
@@ -27,19 +31,19 @@ export function useAccount() {
...
       setIndexerUrl("");
       return;
     }
 
     const url = import.meta.env.PROD
       ? `${import.meta.env.VITE_CARTRIDGE_API_URL}/x/slot/${
-          params.namespace
+          params.slot
         }/graphql`
       : "http://localhost:8080";
```
bal7hazar, line 43: do we need trailing `/graphql` here? I think yes

bal7hazar (changes requested):
LGTM, just this local url, not sure if it is really an issue

bal7hazar (approved):
LGTM

PR #977, opened 2024-11-04
Fix condition to update provider

JunichiSugiura:
The provider was not properly updated by `rpcUrl` search param


bal7hazar (approved):
LGTM

PR #985, opened 2024-11-05
Apply theme preset to Profile

JunichiSugiura:


packages/controller/src/types.ts
```diff
@@ -139,11 +139,6 @@ export type IFrameOptions = {
   theme?: string;
   /** The color mode to use */
   colorMode?: ColorMode;
-  /** Additional configuration options */
-  config?: {
-    /** Preset themes for the controller */
-    presets?: ControllerThemePresets;
```
bal7hazar, line 145: A bit worried about removing this option which is mainly use to work around the fact that theme was not working properly somehow.
And also it provides a good way to customize your controller on the fly

bal7hazar (approved):
Just a comment/question otherwise LGTM

PR #999, opened 2024-11-07
Nest profile routes

JunichiSugiura:
To support relative path with `<Link />` component or `useNavigate` hook

### DIFF
--- a/packages/profile/src/components/trophies/leaderboard.tsx
+++ b/packages/profile/src/components/trophies/leaderboard.tsx
@@ -4,10 +4,9 @@ import {
   SparklesIcon,
   StateIconProps,
 } from "@cartridge/ui-next";
-import { Link, useLocation } from "react-router-dom";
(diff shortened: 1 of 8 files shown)


packages/profile/src/components/trophies/leaderboard.tsx
```diff
@@ -43,15 +42,9 @@ function Row({
...
-  const location = useLocation();
-
-  const path = useMemo(() => {
-    if (self) return location.pathname;
-    return [...location.pathname.split("/"), address].join("/");
-  }, [location.pathname, address, self]);
 
   return (
-    <Link className="flex" to={path}>
+    <Link className="flex" to="..">
```
bal7hazar, line 47: Here we need to add the address in params if the user click on someone except himself (self)

bal7hazar (changes requested):

bal7hazar (approved):
LGTM

PR #1009, opened 2024-11-09
Sign typed data with session

xJonathanLEI:
Contract changes:

- Updated `is_valid_signature` to accept a new signature type with a magic prefix `session-typed-data`, which encodes:

  - pairs of `TypedData`
  - a `SessionToken`

  This allows calling a single unified function for verifying SNIP-12 signatures, regardless of what signer is used.

Account SDK changes:

- Modified `account_sdk::controller::Controller::sign_message` to use session signer if available. The resulting signature follows the new `session-typed-data` format.
- Added a new `hasSessionForMessage` wasm function for checking session presence, similar to the existing `hasSession`.

Keychain changes:

- The `signMessage` API now takes an additional `sync` param for explictly choosing to prompt for signature; otherwise it just uses a session if available.

Controller changes:

- Would now attempt `signMessage` with `sync = false` first to see if the message can be signed headlessly, similar to how `execute` works.


packages/controller/src/types.ts
```diff
@@ -104,6 +104,7 @@ export interface Keychain {
   signMessage(
     typedData: TypedData,
     account: string,
+    sync?: boolean,
```
tarrencev, line 107: hmm I think we need to invert this to make it backward compatible, since there are existing controller libraries in the wild. so undefined / false should have the existing behavior
```suggestion
    async?: boolean,
```

xJonathanLEI: Oh that's true indeed. When designing this I only considered "getting something signed" as the keychain's public contract, not that "some UI is displayed". That's why I didn't consider changing the behavior here a breakage.

I can update to flip the behavior tho.

xJonathanLEI: Updated to flip to default.

tarrencev (approved):

PR #1014, opened 2024-11-11
Fix invalid address copied on profile

xJonathanLEI:
The copied address is incorrectly shortened with `...` in the middle.


bal7hazar (approved):
LGTM

PR #1022, opened 2024-11-12
Fix posthog env for profile

JunichiSugiura:
Profile is a Vite app not Next JS hence the env prefix is different.


bal7hazar (approved):
LGTM

PR #1034, opened 2024-11-15
ERC20 transfer from Profile

JunichiSugiura:


bal7hazar (approved):
LGTM, just few comments

Transfer looks to work, but after clicking Send, the page is empty
image

Also I think the check on the address validity should be done when the Input component lose focus. It seems to only appears when we click on Review send, which is a bit late in the UX perspective I think.
image

It doesnt seem to be locked over the decimal of the ERC20, so I can put 1E-19 ETH when the token actually has 18 decimals.
Also for small number there is a conversion issue
image

PR #1038, opened 2024-11-19
Update token related Torii query

JunichiSugiura:
- Adapt to new `tokenBalances` and `tokenTransfers` query
- Support above query in activity page 
- Fix decimal conversion issue in sending flow


bal7hazar (approved):
LGTM

PR #1066, opened 2024-11-26
Fix initial loading issue with isReady flag

JunichiSugiura:
Torii GQL error was thrown due to `indexerUrl` isn't set initially. Query should be only when `isReady` flag becomes true.


bal7hazar (approved):
LGTM

PR #1074, opened 2024-12-01
Docs: (Readme)

Ocheretovich:
Replaced the link as the previous one was incorrect


tarrencev:
Thanks for the contribution. The existing link is in fact correct, but isn't intended to be visited directly (it is designed to be embedded through an iframe), so nothing renders on the home page.

Ocheretovich:
Then excuse me

PR #1075, opened 2024-12-02
Introducing verified config

JunichiSugiura:


packages/controller/src/controller.ts
```diff
@@ -112,7 +111,7 @@ export default class ControllerProvider extends BaseProvider {
 
     try {
       let response = await this.keychain.connect(
-        this.options.policies || [],
+        this.options.policies || {},
```
tarrencev, line 114: Can we avoid making any backwards incompatible changes to `packages/controller`?

JunichiSugiura:
`packages/keychain/.storybook/preview.tsx` is somehow ignored.

PR #1111, opened 2024-12-06
Session summary updates

tarrencev:


packages/keychain/src/hooks/connection.ts
```diff
@@ -128,8 +128,7 @@ export function useConnectionValue() {
     // Provides backward compatability for Controler <= v0.5.1
     if (themeParam) {
       const decodedPreset = decodeURIComponent(themeParam);
-      const parsedTheme = JSON.parse(decodedPreset) as ControllerTheme;
```
tarrencev: Hm I think we need to keep this unfortunately, since old controllers pass the theme as a json object in the query params

broody: Ah right, putting this in a try/catch

PR #1124, opened 2024-12-09
Feat(profile)/transfer

bal7hazar:


JunichiSugiura (changes requested):
Thanks for refining this flow. Code looks good to me. Just need to test it but not sure how to solve the iframe issue in the preview on this PR. Unless we find a quick solution, maybe we should merge and test in main branch preview with Sepolia account right after. We need to disable the sending button in case when games uses old version of controller without `sendBackward/foward` method tho.

packages/profile/src/components/inventory/token/token.tsx
```diff
@@ -30,10 +30,16 @@ import { formatEther } from "viem";
...
 import { TokenPair } from "@cartridge/utils/api/cartridge";
+import { usePostHog } from "posthog-js/react";
+import { useEffect } from "react";
 
 export function Token() {
   const { address } = useParams<{ address: string }>();
   const location = useLocation();
+  const posthog = usePostHog();
+  useEffect(() => {
+    posthog.setPersonPropertiesForFlags({ address });
```
tarrencev, line 41: can we move this to the connection where we set other posthog metadata?

JunichiSugiura: moved it to `useAccount` hook

broody (approved):
Erc20 transf erworks locally for me, I think should be fine to release and address comments

JunichiSugiura (approved):
LGTM

PR #1153, opened 2024-12-16
refactor: encapsulate `Controller.cartridge`

xJonathanLEI:
The `keychain` codebase currently uses a mix of `Controller` encapsulation and directly accessing `Controller.cartridge`. This is bad and `Controller.cartridge` should just be encapsulated instead to create a unified point of entry to calling anything wasm.

This PR makes the field private and refactors any sites accessing it.


xJonathanLEI:
Superseded by #1155.

PR #1155, opened 2024-12-16
fix: unsynchronized concurrent wasm calls

xJonathanLEI:
Concurrent calls to the wasm module where at least one call borrows `CartridgeAccount` mutably causes `wasm-bindgen`'s runtime borrow checker to throw:

```
recursive use of an object detected which would lead to unsafe aliasing in rust
```

This PR fixes it by:

1. Reducing unnecessary borrowing: A new `CartridgeAccountWithMeta` type is now used for reading fixed values. This type is only ever borrowed immutably, so concurrent accesses are fine.
2. Guarding the inner `controller` value with a Promise-backed `WasmMutex`, and changing all entrypoint signatures to use `&self` instead of `&mut self`.

> [!NOTE]
>
> Using a mutex for all calls, as implemented in this PR, is an overkill. A more ideal implementation would be something like a `RwLock` such that concurrent reads are still allowed.

### DIFF
--- a/packages/keychain/src/pages/index.tsx
+++ b/packages/keychain/src/pages/index.tsx
@@ -12,11 +12,13 @@ import { Upgrade } from "components/connect/Upgrade";
 import { PurchaseCredits } from "components/Funding/PurchaseCredits";
 import { useEffect, useState } from "react";
 import { usePostHog } from "posthog-js/react";
+import { PageLoading } from "components/Loading";
 
 function Home() {
   const { context, controller, error, policies, upgrade } = useConnection();
-  const [hasSessionForPolicies, setHasSessionForPolicies] =
-    useState<boolean>(false);
+  const [hasSessionForPolicies, setHasSessionForPolicies] = useState<
+    boolean | undefined
+  >(undefined);
   const posthog = usePostHog();
 
   useEffect(() => {
@@ -41,7 +43,7 @@ function Home() {
         setHasSessionForPolicies(!!session);
       });
     } else {
-      setHasSessionForPolicies(false);
+      setHasSessionForPolicies(undefined);
     }
   }, [controller, policies]);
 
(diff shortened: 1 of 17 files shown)


packages/keychain/src/pages/index.tsx
```diff
@@ -73,7 +85,7 @@ function Home() {
         return <></>;
       }
 
-      if (controller.session(policies)) {
+      if (hasSessionForPolicies) {
```
tarrencev, line 88: Hmm does this migration to async maintain the same behavior? It seems on initial render it will be false and proceed now. I think we might need a loading state for the intermediate state, when the `useEffect` hasn't been executed

packages/keychain/src/pages/session.tsx
```diff
@@ -125,19 +125,21 @@ export default function Session() {
...
+          expiresAt: String(SESSION_EXPIRATION),
+        });
 
-      return;
-    }
+        return;
+      }
 
-    setIsLoading(false);
+      setIsLoading(false);
```
tarrencev, line 141: Should we move this into the promise resolution handler?

packages/keychain/src/pages/index.tsx
xJonathanLEI: I think you're right that this is _technically_ not the exact same behaviour as the sync version. However, I think under a real world scenario it's _very_ unlikely that users would ever run into the case where the loading state is observable. This is because when `.connect()` is called, it's highly unlikely that the host app would somehow also be blocking the wasm module. After all, `.connect()` is likely their first function call before doing anything real. This means that despite being `async`, the `hasSessionForPolicies` state would instantaneously be set to its desired value.

That said, I can certainly add a loading state. It's just that given the unlikeliness of it being actually observable as discussed above, I'd go for something very primitive. I will test it by artificially delaying the wasm call.

xJonathanLEI: Alright I just pushed a commit adding a loading screen using the existing `<PageLoading />` component:

xJonathanLEI:
Got conflicts. Rebasing now.

xJonathanLEI:
Rebased.

tarrencev (approved):

rsodre:
this is also happening on dojo.js torii client
will open an issue...

PR #1178, opened 2024-12-18
Temporarily allow logins for usernames with periods

broody:
We have some controller accounts that were created with periods but now disallowed for login. This PR allows them to login again to transfer their assets out of the controller.


JunichiSugiura (approved):
LGTM

PR #1191, opened 2024-12-19
✨ ERC20 Transfers features: SNS - Controller - Wallet provider info

bal7hazar:


packages/profile/src/components/inventory/index.tsx
```diff
@@ -1,5 +1,6 @@
 export { Asset, Collection, SendCollection } from "./collection";
-export { Token, SendToken } from "./token";
+export { Token } from "./token";
+export { SendToken } from "./token/send/index";
```
JunichiSugiura, line 3: i think you can omit index

packages/profile/src/components/inventory/token/send/recipient.tsx
```diff
@@ -0,0 +1,200 @@
...
+  setWarning,
+}: {
+  to: string;
+  setTo: (to: string) => void;
+  setWarning: (warning: string) => void;
+}) => {
+  const [focus, setFocus] = useState(false);
+  const [value, setValue] = useState("");
+  const [selectedName, setSelectedName] = useState("");
+  const [selectedWallet, setSelectedWallet] = useState<Wallet | null>(null);
```
JunichiSugiura, line 32: I avoid using `null` as `undefined` complements them often. Not a big deal tho.

packages/profile/src/components/layout/index.tsx
```diff
@@ -74,33 +74,42 @@ type LayoutHeaderProps = {
...
   right?: React.ReactElement;
+  rounded?: boolean;
 };
 
 export function LayoutHeader({
   title,
   description,
   icon,
   right,
+  rounded = false,
```
JunichiSugiura, line 85: I often skip default value as `undefined` is already a falsy value.

packages/profile/src/hooks/starknetid.ts
```diff
@@ -0,0 +1,24 @@
+import { useQuery } from "react-query";
+import { useConnection } from "./context";
+import { useState } from "react";
+
+export const useStarkAddress = ({ name }: { name: string }) => {
+  const { provider } = useConnection();
+  const [error, setError] = useState<string | null>(null);
+
+  const { data, isFetching } = useQuery({
```
JunichiSugiura, line 9: I added swc dep for async fetching logic but you make me realize that react query is just enough 👍

packages/profile/package.json
```diff
@@ -18,6 +18,7 @@
     "@cartridge/utils": "workspace:*",
     "@hookform/resolvers": "^3.9.1",
     "compare-versions": "^6.1.1",
+    "lodash": "^4.17.21",
```
JunichiSugiura, line 21: ECMAScript would never evolve to the point that we no longer need lodash eh? 😭

JunichiSugiura (approved):
LGTM, just some small comments

PR #1207, opened 2024-12-23
Sunset legacy field

JunichiSugiura:


bal7hazar (approved):

tarrencev:
It looks like this includes the input changes but also the toast and CopyAddress migration. Can we land these all separately?

JunichiSugiura:
@tarrencev ah yes, it seems like the commit history tangled when base branch was changed.

JunichiSugiura:
Commit history should be cleaned now.

PR #1216, opened 2024-12-30
Sunset legacy link

JunichiSugiura:


packages/keychain/src/components/ErrorAlert.tsx
```diff
@@ -367,13 +367,14 @@ function StackTraceDisplay({
...
                       <Link
-                        href={getExplorerUrl(
-                          key === "address" ? "contract" : "class",
-                          value as string,
-                        )}
-                        isExternal={isExternalLink}
-                        wordBreak="break-all"
-                        textAlign="left"
+                        to={
+                          getExplorerUrl(
```
tarrencev, line 371: Let's also migrate this to the starknet react explorer getter in a separate PR

tarrencev (approved):

packages/keychain/src/components/ErrorAlert.tsx
tarrencev: Like this: https://github.com/cartridge-gg/controller/pull/1143/files#diff-4b7c03169109fc3f8d54bf8fd431d32d6e87ba9c22904cb5ca051fd34f56545dR18

PR #1219, opened 2024-12-31
Adapt to useExplorer from starknet react

JunichiSugiura:
ref https://github.com/cartridge-gg/controller/pull/1216#discussion_r1899695725

### DIFF
--- a/packages/keychain/src/components/ErrorAlert.tsx
+++ b/packages/keychain/src/components/ErrorAlert.tsx
@@ -19,14 +19,13 @@ import {
   Divider,
 } from "@chakra-ui/react";
 import { motion } from "framer-motion";
-import React, { ReactElement, useEffect, useState } from "react";
+import React, { ReactElement, useCallback, useEffect, useState } from "react";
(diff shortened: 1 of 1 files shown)


packages/keychain/src/components/ErrorAlert.tsx
```diff
@@ -367,13 +350,8 @@ function StackTraceDisplay({
...
                     {key === "address" || key === "class" ? (
                       <Link
-                        to={
-                          getExplorerUrl(
-                            key === "address" ? "contract" : "class",
-                            value as string,
-                          ) as string
-                        }
-                        target={isExternalLink ? "_blank" : undefined}
+                        to={explorer.transaction(value as string)}
```
tarrencev, line 353: hmm i dont think this is a transaction right? previously it was either contract or class

tarrencev (changes requested):

packages/keychain/src/components/ErrorAlert.tsx
JunichiSugiura: oops, my bad

PR #1227, opened 2024-12-31
Set cover image via css variable

JunichiSugiura:
To get rid of `useColorMode` from Chakra.


JunichiSugiura:
It seems working find on Keychain app but dark cover image isn't applied to Storybook somehow.

This trick isn't working on Storybook I guess.
https://github.com/cartridge-gg/controller/blob/fc7992f9486aa0892227ca304c92039e58f97b24/packages/ui-next/src/hooks.ts#L14-L16

PR #1242, opened 2025-01-07
Add chain-id to account storage key

broody:
This will help with multi accounts/networks support by using the chain id as an additional key to `account` in local storage.


tarrencev (approved):
Will this require everyone to recreate sessions?

broody:
> Will this require everyone to recreate sessions?

Yeah, should auto prompt

PR #1243, opened 2025-01-07
Disconnect only clears active

broody:
Currently disconnect clears out everything in storage, this PR changes to just clearing out active. On reconnect, we reuse previous valid session instead of creating new one.


tarrencev:
I think we want the behavior where everything is cleared. Since if you're using a shared device, you'll unintentionally leave a valid session on the device when disconnecting

PR #1260, opened 2025-01-08
🔧 Turn example into https and profile handles wasm cfg

bal7hazar:


examples/next/package.json
```diff
@@ -3,7 +3,7 @@
   "private": true,
   "version": "0.5.8",
   "scripts": {
-    "dev": "next dev -p 3002",
+    "dev": "next dev -p 3002 --experimental-https",
```
bal7hazar, line 6: Turning the example to HTTPS so we can actually test the prod directly and avoid this:
image

broody (approved):

PR #1272, opened 2025-01-13
Credits funding for slot

broody:


steebchen (approved):

packages/keychain/src/components/funding/PurchaseCredits.tsx
```diff
@@ -130,13 +113,14 @@ export function PurchaseCredits({ onBack }: PurchaseCreditsProps) {
...
       onBack={state === PurchaseState.SELECTION ? onBack : undefined}
+      hideNetwork
     >
       <Content className="gap-6">
         <Balance showBalances={["credits"]} />
         <ErrorAlert
           variant=""
           title="WHAT ARE CREDITS"
-          description="Credits can be used to play games. They are not tokens and cannot be transferred or refunded."
+          description="Credits can be used to play games or pay for slot deployments. They are not tokens and cannot be transferred or refunded."
```
tarrencev, line 123: Hmm could we make this conditional based on the users entrypoint? So we only mention slot when coming from the slot funding flow?

tarrencev:
Please run `scripts/update-storybook-snapshot.sh` to update the snapshots

PR #1310, opened 2025-01-21
Only shim fetch-cookie + pako in prod

tarrencev:


bal7hazar (approved):
Tested, it solves the issue

PR #1323, opened 2025-01-24
refactor: send raw json when callback uri

Larkooo:


packages/keychain/src/components/session.tsx
```diff
@@ -75,7 +71,7 @@ export function Session() {
           }
 
           const res = await fetch(url, {
-            body: encodedResponse,
+            body: JSON.stringify(response),
```
tarrencev, line 74: Will this break slot?

Larkooo: i think slot goes through another flow no? the oauth stuff

tarrencev: ah yeah what about sozo though? doesn't it use this flow to create a cli controller?

Larkooo: ah ok yes that will need to be changed to use the decode the raw json directly. i just wonder if the flow still works because the rev that slot uses for the account_sdk is outdated and when i used it, i had deserialization errors when initializing and using a session account. @glihm do you know if it still works fine?

PR #1338, opened 2025-01-28
Bust cache if controller serde issue

broody:


packages/account_sdk/src/tests/account/storage.rs
```diff
@@ -0,0 +1,42 @@
+#[cfg(test)]
```
tarrencev, line 1: Lets move this to a `_test.rs` file next to storage rather than this tests folder

packages/account_sdk/src/storage/filestorage.rs
```diff
@@ -44,12 +44,16 @@ impl FileSystemBackend {
 #[async_trait]
 impl StorageBackend for FileSystemBackend {
     fn set(&mut self, key: &str, value: &StorageValue) -> Result<(), StorageError> {
+        let serialized = serde_json::to_string(value)?;
+        self.set_serialized(key, &serialized)
+    }
+
+    fn set_serialized(&mut self, key: &str, value: &str) -> Result<(), StorageError> {
```
tarrencev, line 51: Hmm I'm not sure I understand why this additional method is necessary?

broody: So actually only used in the test rn, needed a way to set a corrupt value into storage to test triggering deserialization error. I guess we can also make this `pub` https://github.com/cartridge-gg/controller/blob/49866a7faa7cde855085ff6f39263d27b790eaa4/packages/account_sdk/src/storage/inmemory.rs#L8

PR #1346, opened 2025-01-30
Add dominion preset

ccbbccbb:


tarrencev:
Hey @ccbbccbb this should be contributed here in the presets repo, following these docs: https://docs.cartridge.gg/controller/presets

Is there some old docs that pointed you in this direction?

PR #1372, opened 2025-02-06
Added individual toggle functionality for create session

tarrencev:


packages/keychain/src/components/connect/CreateSession.tsx
```diff
@@ -41,6 +42,7 @@ export function CreateSession({
       Object.keys(policies.contracts).forEach((address) => {
         if (policies.contracts![address]) {
           policies.contracts![address].methods.forEach((method) => {
+            method.id = crypto.randomUUID();
```
tarrencev, line 45: hm i think you can use something like `${address}-${method}`

PR #1377, opened 2025-02-06
✨ Add recipient and create account modules and refactor colors and create color storybook

bal7hazar:
image
image
image


tarrencev:
image

Can we hide the `x` if the input is empty? Additionally, lets make the placeholder `Recipient Address or Username`

image
Can we improve the wrapping here?

bal7hazar:
> Can we hide the `x` if the input is empty? Additionally, lets make the placeholder `Recipient Address or Username`

image

> Can we improve the wrapping here?

I have turned the font from `sans` to `mono` to avoid unexpected wrapping
A normal formatted Starknet address should take exactly 2 lines

image
image
image

And more for unexpected longer string

image

tarrencev:
@bal7hazar one last thing, it seems like all the text got shifted up 1px

image

tarrencev (approved):

PR #1407, opened 2025-02-13
Fix policy methods not showing on register session

nrmnqdds:
before:
image

after:
CleanShot 2025-02-13 at 23 00 13@2x


packages/keychain/src/components/ExecutionContainer.tsx
```diff
@@ -56,6 +58,8 @@ export function ExecutionContainer({
     "execute",
   );
 
+  const { isEditable, onToggleEditable } = useCreateSession();
```
tarrencev, line 61: Imo it would be better to pass the right item element as a prop to execution container than to access `useCreateSession` here, so we don't couple session creation at this low level

packages/keychain/src/components/ExecutionContainer.tsx
```diff
@@ -1,22 +1,22 @@
-import { useState, useCallback, useEffect, useRef } from "react";
+import { ControllerErrorAlert, ErrorAlert } from "@/components/ErrorAlert";
+import { useConnection } from "@/hooks/connection";
+import type { ControllerError } from "@/utils/connection";
+import { parseControllerError } from "@/utils/connection/execute";
+import { ErrorCode } from "@cartridge/account-wasm/controller";
```
tarrencev, line 5: @nrmnqdds I think we can remove all of these now

packages/keychain/src/components/connect/RegisterSession.tsx
```diff
@@ -28,17 +35,79 @@ export function RegisterSession({
   const [transactions, setTransactions] = useState<Call[] | undefined>(
     undefined,
   );
+  const [isEditable, setIsEditable] = useState(false);
+  const [policyState, setPolicyState] = useState<ParsedSessionPolicies>(() => {
```
tarrencev, line 39: can we move most of this into the context / a hook? so the implementation isn't duplicated between here and create session?

PR #1408, opened 2025-02-13
Implement wildcard session supports

tarrencev:


packages/keychain/src/components/connect/create/useCreateController.ts
```diff
@@ -143,13 +142,21 @@ export function useCreateController({
...
+              chainId!,
+              rpcUrl!,
               username,
               controllerNode.constructorCalldata[0],
               controllerNode.address,
               credentialId,
               publicKey,
             );
+
+            await controller.login(NOW + DEFAULT_SESSION_DURATION);
```
broody, line 156: If I understand right, this creates a wildcard session, but should this be done for every login? Even if policies are defined?

broody:
Maybe I'm missing something from this flow, it seems like we no longer call `create_session` on controller at all now https://github.com/cartridge-gg/controller/pull/1408/files#diff-831d2843befb8f6b0c34773f63f9ba3b49deac571c21364b451372114e0093c2L178

So when I run the next example after login/signup every txn requires a webauthn signature, no controller prompt just the native webauthn prompt

broody (approved):
LGTM

PR #1418, opened 2025-02-19
Port tokens context logic from keychain to utils in order to share with profile

JunichiSugiura:


packages/keychain/__image_snapshots__/components-deploycontroller--default-chromium.png
JunichiSugiura, line 1: this change is due to the new mocked data. i'll create another pr to show all cases of `<DeployController />` component.

PR #1422, opened 2025-02-21
Pass SNAPSHOT env directly to serve command

JunichiSugiura:
`docker run -e SNAPSHOT=1` does not seems to be working so passing the flag directly to pnpm script which serves storybook.


JunichiSugiura:
Should be working this time as the angle of spinner matches to svg export from Figma

Screenshot 2025-02-21 at 14 18 31

Screenshot 2025-02-21 at 14 17 33

PR #1430, opened 2025-02-25
Cleanup dependencies

JunichiSugiura:


packages/ui-next/package.json
```diff
@@ -71,6 +71,7 @@
   },
   "devDependencies": {
     "@cartridge/eslint": "workspace:^",
+    "@cartridge/presets": "catalog:",
```
tarrencev, line 74: is this needed?

packages/utils/package.json
```diff
@@ -28,6 +28,7 @@
     "viem": "catalog:"
   },
   "devDependencies": {
+    "@cartridge/presets": "catalog:",
```
tarrencev, line 31: is this needed?

package.json
```diff
@@ -30,9 +30,6 @@
     "test:storybook": "pnpm turbo build:deps test:storybook",
     "test:storybook:update": "pnpm turbo build:deps test:storybook:update"
   },
-  "dependencies": {
-    "@cartridge/presets": "github:cartridge-gg/presets#c7e6fb7"
```
tarrencev, line 34: we need to make sure this change works with the presets update ci

packages/ui-next/package.json
JunichiSugiura: This is used in storybook for mainly layout thus devDependencies.

https://github.com/cartridge-gg/controller/blob/407797509c7c569bbde474676dede08d2b4af4d3/packages/ui-next/.storybook/preview.tsx#L10

packages/utils/package.json
JunichiSugiura: This is used to pass ERC20 icons to ERC20 class. In the future, it should be provided via gateway API.

package.json
JunichiSugiura: Good point !

PR #1440, opened 2025-02-27
✨ Add thumbnails components

bal7hazar:


packages/controller/src/iframe/profile.ts
```diff
@@ -48,6 +49,16 @@ export class ProfileIFrame extends IFrame<Profile> {
       );
     }
 
+    if (policies?.contracts) {
```
tarrencev, line 52: Hmm curious is this intended to be part of this PR?

packages/controller/src/types.ts
```diff
@@ -183,6 +183,8 @@ export type ProfileOptions = IFrameOptions & {
   namespace?: string;
   /** The tokens to be listed on Inventory modal */
   tokens?: Tokens;
+  /** The policies to use for the profile */
+  policies?: SessionPolicies;
```
tarrencev, line 187: Is this intended to be part of this PR?

packages/ui-next/src/assets/index.ts
```diff
@@ -0,0 +1,2 @@
+export const PLACEHOLDER =
```
tarrencev, line 1: Not necessary for this PR, but we should probably move towards including icons as a static asset, so they don't bloat the bundle

tarrencev (approved):

packages/controller/src/types.ts
bal7hazar: Yes it does because we need to know the entrypoints defined in the policies inside the profile iframe to then query the corresponding transactions from toriis

packages/ui-next/src/assets/index.ts
bal7hazar: For some reason I was not able to manage svg in the ui-next package, either not supported or not exported, I tried hard and for so long but there is probably a specific config somewhere avoiding it that I didnt found, I just ended it with the b64 version of the file to unlock the situation

packages/controller/src/iframe/profile.ts
bal7hazar: This will be temporary until we have something better with torii, but ultimately we will need correspondance between entrypoint hash and plain text (if it cannot be done through a RPC call to get the abi at the indexation time)

PR #1455, opened 2025-03-06
Feat/add devnet chainid

MartianGreed:
## Summary by CodeRabbit

- **New Features**
	- Enhanced event management by introducing capabilities to retrieve clear event identifiers and names.
	- Streamlined event processing with a more flexible and unified approach to handling various event types.
	- Improved event tracking for both account-related and token-related activities.

### DIFF
--- a/packages/controller/src/account.ts
+++ b/packages/controller/src/account.ts
@@ -54,8 +54,6 @@ class ControllerAccount extends WalletAccount {
    * @returns response from addTransaction
    */
   async execute(calls: AllowArray<Call>): Promise<InvokeFunctionResponse> {
-    console.log("account->execute", calls, this);
-
     calls = toArray(calls);
 
     return new Promise(async (resolve, reject) => {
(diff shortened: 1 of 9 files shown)


glihm:
Could be merged after #1454 for a cleaner diff as we discussed. 👍 
Bump review cc @tarrencev @broody @bal7hazar

packages/controller/src/account.ts
```diff
@@ -54,6 +54,8 @@ class ControllerAccount extends WalletAccount {
    * @returns response from addTransaction
    */
   async execute(calls: AllowArray<Call>): Promise<InvokeFunctionResponse> {
+    console.log("account->execute", calls, this);
```
broody, line 57: Remove debug

packages/controller/src/controller.ts
```diff
@@ -298,6 +298,7 @@ export default class ControllerProvider extends BaseProvider {
         `Chain not found: ${shortString.decodeShortString(this.selectedChain)}. Available chains: ${availableChains.join(", ")}`,
       );
     }
+    console.log(chain, this.selectedChain);
```
broody, line 301: Remove debug

broody (changes requested):
Massive unlock! Bravo @MartianGreed @glihm 👏

Plz also run `pnpm format`

broody:
Added a small modification - refactored `parseChainId` to retrieve chain id via rpc call if localhost so we don' t need to hardcode it

PR #1458, opened 2025-03-07
chore: update @cartridge/presets

tarrencev:
Updates @cartridge/presets dependency to point to commit b7d379e


bal7hazar (approved):

tarrencev:
@bal7hazar this requires an update to keychain to use the new chain specific policies schema before we can merge

PR #1475, opened 2025-03-13
🔥 Remove unsupported svg rules

bal7hazar:


tarrencev:
This seems to break some of the svg rendering (the `!` in the caution sign)

bal7hazar:
> This seems to break some of the svg rendering (the `!` in the caution sign)

Ok should be good now

tarrencev (approved):

PR #1477, opened 2025-03-13
Add multi-wallet support

broody:
This PR allows external wallets like phantom, metamask, argent to communicate with keychain. Due to iframe restrictions we can't directly access these wallets' window objects, so we'll add some additional methods to keychain iframe message channel to handle wallet actions.


packages/controller/package.json
```diff
@@ -60,17 +60,19 @@
...
     "@cartridge/penpal": "catalog:",
+    "@metamask/sdk": "^0.32.1",
+    "@phantom/wallet-sdk": "^0.0.23",
     "@starknet-io/types-js": "catalog:",
     "@telegram-apps/sdk": "^2.4.0",
-    "base64url": "catalog:",
     "cbor-x": "^1.5.0",
-    "fast-deep-equal": "catalog:"
+    "fast-deep-equal": "catalog:",
+    "starknetkit": "^2.6.1"
```
tarrencev, line 75: I wonder if most of these should be peer deps since the calling app will likely have them already included

tarrencev (approved):

PR #1479, opened 2025-03-14
Update Controller cairo

tarrencev:


packages/account_sdk/artifacts/metadata.json
```diff
@@ -1,8 +1,8 @@
 {
   "controllers": {
     "latest": {
-      "casm_hash": "0x2fcd4b347622fd2dbe973ad873666f8925b3f244120793014248b4fd9670bd9",
-      "class_hash": "0x743c83c41ce99ad470aa308823f417b2141e02e04571f5c0004e743556e7faf"
+      "casm_hash": "0x462d4f81293ffa97dc725e28179dfa8e7160b4b92f00c8dc2acd248bcd9cc01",
+      "class_hash": "0x35fe40d30fd6937d972492ed09f9f409973fa61b3cc2fe821c3ebb517c1c44d"
```
glihm, line 5: Is this class hash supposed to be `1.0.10`? Or latest doesn't specifically match an existing version? If so, will need to update some dependencies in Dojo to take this new behavior in account.

tarrencev: The latest is just for development right now, it isn't a live class. Eventually we'll promote it to `1.0.10` but it will change a few times before then

PR #1489, opened 2025-03-21
Initial Sign in with Solana implementation

tarrencev:


edisontim (commented):
Is there any way I can do an e2e test of this already? Or that thing missing on the API side is going to block me from doing that?

PR #1504, opened 2025-03-25
Use vite build for controller pkg

broody:
This is needed to as controller pkg deps on more external sdk for wallet integration. Keychain is built using vite, but becomes incompatible with controller pkg built using tsup without polyfill. 

This PR splits the controller build between browser and node - browser will use `vite` and node remains using `tsup`

### DIFF
--- a/packages/controller/package.json
+++ b/packages/controller/package.json
@@ -28,14 +28,6 @@
       "types": "./dist/node/index.d.ts",
       "import": "./dist/node/index.js",
       "require": "./dist/node/index.cjs"
-    },
-    "./provider": {
-      "types": "./dist/provider/index.d.ts",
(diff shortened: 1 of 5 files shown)


packages/controller/package.json
```diff
@@ -16,52 +18,21 @@
...
-      "require": "./dist/session/index.cjs"
+      "types": "./dist/session.d.ts",
+      "import": "./dist/session.js"
     },
     "./session/node": {
       "types": "./dist/node/index.d.ts",
       "import": "./dist/node/index.js",
       "require": "./dist/node/index.cjs"
-    },
-    "./provider": {
```
broody, line 32: Removing these as they were pointing to non-existent dist files

PR #1517, opened 2025-04-01
Fix logout behavior

nrmnqdds:


broody (changes requested):

packages/controller/src/controller.ts
```diff
@@ -51,6 +51,15 @@ export default class ControllerProvider extends BaseProvider {
 
     if (typeof window !== "undefined") {
       (window as any).starknet_controller = this;
+
+      // Add event listener for messages from iframes
+      window.addEventListener("message", (event) => {
```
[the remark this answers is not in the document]
nrmnqdds: 93510ac

packages/keychain/src/hooks/connection.ts
```diff
@@ -236,6 +238,18 @@ export function useConnectionValue() {
               message: "User logged out",
             });
           }
+
+          const connection = connectToParent<ParentMethods>({
```
broody, line 242: What's the purpose of this after logout?

broody (approved):
LGTM, just had a question about calling `connectToParent` after logout

packages/keychain/src/hooks/connection.ts
broody: I see, we can just reuse `parent` instead of making another connection, updated. Feel free to merge if u see no other issues, thx!

PR #1538, opened 2025-04-10
Wallet Bridge interface

tarrencev:


packages/account_sdk/src/signers/external.rs
```diff
@@ -0,0 +1,90 @@
...
+use crate::signers::SignError;
+use crate::utils::js::extract_error_message;
+use js_sys::Promise;
+use serde::{Deserialize, Serialize};
+use wasm_bindgen::prelude::*;
+use wasm_bindgen_futures::JsFuture;
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+#[serde(rename_all = "lowercase")]
+pub enum ExternalWalletType {
```
tarrencev, line 12: @edisontim @broody i'm not sure if we want the rust to know which eip191 / solana etc wallet we are using for the signing, or if we just want it to naively say "sign with an eip191 signer" and manage that on the js side, or if we should instead use the signer address to identify the appropriate wallet to use

i think the signer address might be the right way to go since that is how the contracts identify the signer

broody (approved):

broody (approved):

PR #1563, opened 2025-04-18
feat: Almost e2e social signup

edisontim:


packages/controller/src/iframe/base.ts
```diff
@@ -45,6 +45,7 @@ export class IFrame<CallSender extends {}> implements Modal {
     iframe.style.border = "none";
     iframe.sandbox.add("allow-forms");
     iframe.sandbox.add("allow-popups");
+    iframe.sandbox.add("allow-popups-to-escape-sandbox");
```
edisontim, line 48: Allows the popup created from within the keychain iframe to be sandboxed but still execute javascript and create its own popups (goal of the popup: authenticate to Discord through Auth0). 
I'm not even sure that it could cause a security issue, it might? if Auth0 or Discord are compromised but how likely is that (famous last words)

packages/controller/src/node/account.ts
edisontim, line 1: Is account.ts from node and from session exactly the same and can’t one of them be removed?

package.json
```diff
@@ -12,7 +12,7 @@
     "e2e:ui": "turbo build:deps dev e2e:ui",
     "lint": "turbo lint format:check",
     "format": "turbo format lint:fix",
-    "clean": "git clean -xdf && pnpm store prune",
+    "clean": "pnpm store prune",
```
edisontim, line 15: @JunichiSugiura is this ok for you if I remove this? Got some untracked files deleted haha

packages/keychain/src/hooks/wallets.tsx
```diff
@@ -149,36 +150,63 @@ export const useWallets = (): WalletsContextValue => {
...
    * @param identifier - The identifier (e.g., address) of the wallet.
    * @param message - The message hex string to sign.
    * @returns Promise resolving to the signature hex string.
    * @throws Error if signing fails or the connection isn't ready.
    */
-  async signMessage(identifier: string, message: string): Promise<string> {
+  async signMessage(
+    identifier: string,
+    message: string,
+  ): Promise<ExternalWalletResponse> {
```
edisontim, line 189: Think it’s better here to return the full response of the embedded/bridge instead of just the signature. Imo it contains valuable information such as the type of wallet, address, etc.

packages/account-wasm/src/types/signer.rs
```diff
@@ -85,6 +88,18 @@ impl TryFrom<StarknetSigner> for SigningKey {
...
 }
 
+impl TryFrom<Eip191Signer> for account_sdk::signers::eip191::Eip191Signer {
+    type Error = EncodingError;
+
+    fn try_from(eip191: Eip191Signer) -> Result<Self, Self::Error> {
+        Ok(Self {
+            address: EthAddress::from_str(&eip191.address).unwrap(),
+            #[cfg(not(target_arch = "wasm32"))]
+            signing_key: alloy_signer::k256::ecdsa::SigningKey::random(&mut rand::rngs::OsRng),
```
edisontim, line 98: Leaving this as random value for now @tarrencev, not sure what needs to be done here yet

packages/keychain/src/components/connect/create/useCreateController.ts
```diff
@@ -96,16 +104,62 @@ export function useCreateController({
...
+      //     session: {
+      //       expiresAt: loginResponse.session.expiresAt,
+      //       allowedPoliciesRoot: loginResponse.session.policies,
+      //       metadataHash: loginResponse.session.metadataHash,
+      //       sessionKeyGuid: loginResponse.session.sessionKeyGuid,
+      //       guardianKeyGuid: loginResponse.session.guardianKeyGuid,
+      //       authorization: loginResponse.authorization,
+      //     },
+      //   });
+      //   console.log(`Registered controller with res: ${res}`);
```
tarrencev, line 162: Remove?

PR #1586, opened 2025-04-25
Display token transactions on profile

nrmnqdds:


nrmnqdds:
image

bal7hazar (approved):
This is nice, I would recommend to split the Token page into components inside ui-next with the storybooks, one for token Header and one for Details at least.

PR #1589, opened 2025-04-28
Options to change transfer token

nrmnqdds:


packages/profile/src/components/inventory/token/send/index.tsx
```diff
@@ -38,27 +45,47 @@ export function SendToken() {
   const [toError, setToError] = useState<Error | undefined>();
   const [submitted, setSubmitted] = useState(false);
   const [loading, setLoading] = useState(false);
+  const [selectedToken, setSelectedToken] = useState<Token | undefined>(token);
   const disabled = useMemo(() => {
     return !!toError || !!amountError || (!validated && !!warning);
   }, [validated, warning, amountError, toError]);
 
+  const filteredAllTokens: Token[] = useMemo(() => {
```
bal7hazar, line 53: I think this filtering could take place right after the source query, in `@src/hooks/token/useTokens`, wdyt?

nrmnqdds: Resolved #1586

bal7hazar:
@nrmnqdds 
There are a lot of small diff vs the target, let's do the following process here:
- Create the core components and storybooks, checking the figma I would expect 3 core components (missing) and 1 macro component (which is the one you did):
  - image
  - image
  - image
  - image
-Validate those with @clicksave
- Review the logic in profile with @bal7hazar (which is already good actually)

bal7hazar (approved):

PR #1595, opened 2025-04-29
Starterpack dynamic supply

broody:


clicksave:
Nice work.

1. felt that the size of the font inside of the "bookmark" tag was a bit too small, so I updated the mocks accordingly  

image

https://www.figma.com/design/OSbzNkyR9UYV4Erz9oF9JJ/%F0%9F%95%B9%EF%B8%8F-Controller?node-id=9817-42808&t=TRNL55yJPIHmZlxd-1

2. Just want to confirm, we're going to be using the theme color for this tag. if theres no theme it should use the default cartridge yellow

image

3. Looks like we're still have that double spacing issue in a few places 

image

Gap here should change from 48px -> 24px

image

4. Column gap should change from 12px -> 16px

image

PR #1617, opened 2025-05-04
ERC20 review transaction page

nrmnqdds:


bal7hazar (changes requested):
Some diff from the design:
Implementation | Target | Comment
:-:|:-:|:-:
image |  image | Extra origin info
image | image | Missing username
image | image | Gap Y diff (consider to not user the inner header in LayoutHeader)
image || Remove ability to select (aka `select-none`)
image | image | Network info row is too tall
image | image | Sending Lords result in a crash (I sent you some to test)

packages/keychain/src/hooks/account.ts
```diff
@@ -238,3 +239,11 @@ export function useAccount() {
     };
   }, [controller]);
 }
+
+export function useUsername(address: string) {
```
bal7hazar, line 243: It doesnt cover other name providers, such as `.stark`
Maybe acceptable until we merge both iframes, need to validate it with @clicksave 

| Profile | Keychain |
|:-:|:-:|
image | image

bal7hazar (changes requested):
Good progression on token transfer, it works well.
However it crashes for anything else than a ERC20 transfer it seems, trying here with a NFT transfer. The confirmation page has been made too ERC20 transfer oriented I think.

image

nrmnqdds:
This PR also resolves C7E-689

PR #1621, opened 2025-05-05
feat: New logins release

edisontim:


packages/account_sdk/src/signers/eip191.rs
```diff
@@ -57,21 +57,16 @@ impl Eip191Signer {
...
-        signature: Signature<Secp256k1>,
+        mut signature: Signature<Secp256k1>,
         y_parity: bool,
     ) -> SignerSignature {
-        // If signature is normalized (s value changed), we need to flip the y_parity
-        let (signature, y_parity) = if let Some(normalized) = signature.normalize_s() {
-            (normalized, !y_parity)
-        } else {
-            (signature, y_parity)
-        };
```
tarrencev, line 68: Do we still want to revert this?

packages/keychain/src/components/connect/buttons/change-wallet.tsx
```diff
@@ -18,80 +21,122 @@ const OPTIONS: Partial<
...
+export function ChangeWallet({
+  validation,
+  changeWallet: externalChangeWallet,
+  setChangeWallet,
+}: ChangeWalletProps) {
   const { wallets } = useWallets();
 
+  useEffect(() => {
+    setChangeWallet(false);
+  }, [validation.status]);
```
tarrencev, line 69: Hmm this seems odd

edisontim: This is to force the change wallet component to disappear if the user starts typing anything else. Granted I think I could put the `validation.signer` in there to make it clearer but the effect is the same

packages/account_sdk/src/signers/eip191.rs
edisontim: Going to check if it still works 100% of the time without it

broody (approved):
LGTM! Also didn't see any compat issues.

Question - I understand we use `useFeature` hook currently to enable this. When this feature is broadly release, is there a way to disable it if the client's controller pkg is less than the version needed?

packages/account_sdk/src/signers/eip191.rs
edisontim: Seems like we can keep the normalization yeah

edisontim:
> LGTM! Also didn't see any compat issues.
> 
> Question - I understand we use `useFeature` hook currently to enable this. When this feature is broadly release, is there a way to disable it if the client's controller pkg is less than the version needed?

I'm looking for a way to do that right now, any chance you have such a pattern already implemented?

edisontim:
I'm also seeing one minor difference with a Passkey controller created on the example and on local API: 
- On passkey I have a few STRKs that I can use for transfers
- With wallet signer I can only transfer 0 STRKs, i.e. I don't have any in the wallet

PR #1640, opened 2025-05-07
feat: Add possibility to choose signup options in ControllerConnector

edisontim:
https://linear.app/cartridge/issue/C7E-686/support-configurable-sign-up-options

### DIFF
--- a/packages/keychain/src/utils/connection/index.ts
+++ b/packages/keychain/src/utils/connection/index.ts
@@ -12,15 +12,18 @@ import { deployFactory } from "./deploy";
 import { openSettingsFactory } from "./settings";
 import { normalize } from "@cartridge/utils";
 import { switchChain } from "./switchChain";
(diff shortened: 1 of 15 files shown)


packages/keychain/src/utils/connection/index.ts
```diff
@@ -12,22 +12,27 @@ import { deployFactory } from "./deploy";
...
   setController: (controller?: Controller) => void;
+  setConfigSignupOptions: (options: AuthOption[] | undefined) => void;
 }) {
   return connectToParent<ParentMethods>({
     methods: {
       connect: normalize(
         connect({
           setRpcUrl,
           setContext,
+          setConfigSignupOptions: (options: AuthOptions | undefined) =>
```
broody, line 34: Small nit - can the method signature above be `options?: AuthOptions`? So we can just pass the name here

broody (approved):
LGTM!

PR #1643, opened 2025-05-08
chore: merge utils and ui-next into ui

MartianGreed:
Merge utils and ui-next packages into `ui`

Main changes: 

- `@cartridge/utils` now available under `@cartridge/ui/utils`
- many packages rename to make it fit properly

Once, this will be merged internally to controller, we could just move the whole package out to [cartridge-gg/ui](https://github.com/cartridge-gg/ui) and it should be pretty transparent to transition to


bal7hazar (approved):
LGTM

PR #1651, opened 2025-05-09
feat: updated ui for qr

ayushtom:
This PR tackles -

- dismissal of tooltip overlay which wasn't working on mobile
- fixing the dialog component to have a default close button behaviour
- updating qr code component to be more flexible with images and data url for use across codebase [can be incrementally updated depending on use]


bal7hazar (changes requested):
The fix works correctly but there are some gap with the target (figma)

| Target | Implementation | Comment |
|:-:|:-:|:-:|
| image | image | The button is not at the right position, a bit too much at the bottom, the padding on top of right of it should be the same |
| image | image | Blur too strong, close button not at the right position, QR code design is different (border space at least, dont know if it is customizable) |
| | image | Make is not selectable please |

bal7hazar (approved):

PR #1652, opened 2025-05-09
feat: remove ui package

MartianGreed:


packages/keychain/vite.config.ts
```diff
@@ -81,5 +81,10 @@ export default defineConfig(({ mode }) => ({
     environment: "jsdom",
     setupFiles: ["./src/test/setup.ts"],
     globals: true,
+    server: {
+      deps: {
+        inline: ["@cartridge/ui"],
+      },
+    },
```
tarrencev, line 88: What does this do?

tarrencev (approved):

packages/keychain/vite.config.ts
MartianGreed: it's for vitest. it doesn't resolves es modules and needs cjs to work. Doing this, we can use @cartridge/ui without having to create separate cjs modules just for testing

PR #1663, opened 2025-05-10
Fix Phantom wallet not being found

broody:
image


edisontim (approved):
Lgtm! Have you tested that account signup and login still work with metamask and rabby?

PR #1700, opened 2025-05-15
Fix/several minor fixes

bal7hazar:
- Added Credits to the Inventory (such as it is in Arcade)
- Added the ability to unfold (starts at 3 and more to not break the first NFT row)
- Rework the Credit page which was outdated, it matches the design now (except credits $value in the header, which is outdated in the design iirc)
- By default only Credits and STRK, can be overridden with the `tokens: { erc20: ("eth" | "strk" | "lords" | "usdc" | "usdt")[] }` Profile option
- We use our internal toriis to get ETH, STRK, USDC/T and LORDS, there is still a fetch balance for all tokens remaining (but a fetch every 30s instead of 3s and disabled when the iframe is closed)
- Fixed Chaos Surfers images

### DIFF
--- a/packages/profile/src/hooks/token.mock.ts
+++ b/packages/profile/src/hooks/token.mock.ts
@@ -6,7 +6,7 @@ export * from "./token";
 
 export const credits = {
   balance: {
-    amount: 1,
+    amount: 1.234567,
     value: 0,
     change: 0,
   },
@@ -15,7 +15,7 @@ export const credits = {
     name: "Credits",
     symbol: "CREDITS",
     image: "https://static.cartridge.gg/presets/credit/icon.svg",
-    decimals: 18,
+    decimals: 6,
   },
 };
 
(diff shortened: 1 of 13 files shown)


packages/profile/src/hooks/token.mock.ts
```diff
@@ -4,8 +4,24 @@ import { UseTokensResponse, UseTokenResponse, Token } from "./token";
...
+    amount: 1,
+    value: 0,
+    change: 0,
+  },
+  metadata: {
+    address: "credits",
+    name: "Credits",
+    symbol: "CREDITS",
+    image: "https://static.cartridge.gg/presets/credit/icon.svg",
+    decimals: 18,
```
broody, line 18: Credits is 6 decimals

packages/profile/src/hooks/token.ts
```diff
@@ -160,19 +167,56 @@ export function useBalances(accountAddress?: string): UseBalancesResponse {
...
+    return {
+      balance: {
+        amount: Number(creditBalance.balance.value),
+        value: 0,
+        change: 0,
+      },
+      metadata: {
+        name: "Credits",
+        symbol: "Credits",
+        decimals: 0,
```
broody, line 196: Credits is 6 decimals

broody (approved):
LGTM! Just a note bout credit decimals.

PR #1710, opened 2025-05-16
feat: add qrcode display on keychain

ayushtom:
The qr code component needed to be shown on keychain too which was not the case. This pr fixes that


bal7hazar (approved):
image
Tested

PR #1726, opened 2025-05-31
Fix blank screen on manual execute after update session

broody:
During manual execution flow, if the session is expired we first show the update session dialog. However, we're losing original context by calling `execute` again when user hits update. This results in a blank screen. This is highly reproducible with pistols.gg

Fix is to use the original `execute` context. But I'm not sure why we were doing it the other way in the first place, perhaps some edge case I'm not aware of cc @tarrencev


tarrencev:
Ah I think the case this might account for is a paymastered txn after session update

tarrencev (approved):

PR #1754, opened 2025-06-13
feat: Re-apply create sessions

edisontim:
A broken dependency introduced regression in the precendent create sessions PR. This fixes it and introduces CI to catch these kind of issues


nrmnqdds (approved):
lgtm

PR #1781, opened 2025-06-23
fix: decimals for credits balance

bal7hazar:


packages/profile/src/hooks/token.ts
```diff
@@ -186,7 +186,7 @@ export function useTokens(accountAddress?: string): UseTokensResponse {
   const credits: Token = useMemo(() => {
     return {
       balance: {
-        amount: Number(creditBalance.balance.value),
+        amount: Number(creditBalance.balance.value) / 10 ** 6,
```
steebchen, line 189: `/ 1e6` might be easier to read/write but both works

PR #1797, opened 2025-06-26
fix: controller version

edisontim:


packages/controller/src/controller.ts
```diff
@@ -189,7 +190,6 @@ export default class ControllerProvider extends BaseProvider {
             : this.options.policies || {},
         this.rpcUrl(),
         this.options.signupOptions,
-        version,
```
edisontim, line 192: @claude will this be a issue? The keychain (which runs in an iFrame) expects one less argument but the controller might not be updated to the right version and might still pass the version in the connect function

edisontim: @claude Hello ????

edisontim: @claude is it really an issue to have an extra parameter?

PR #1843, opened 2025-07-19
Fix incorrect update session request

broody:
Execute context was using stale session detection variable, so after login if transaction call was not part of policies it would request user to unnecessarily update session instead of showing transaction confirmation modal.

### DIFF
--- a/packages/keychain/src/components/connect/CreateSession.tsx
+++ b/packages/keychain/src/components/connect/CreateSession.tsx
@@ -32,10 +32,12 @@ const requiredPolicies: Array<ContractType> = ["VRF"];
 export function CreateSession({
   policies,
   onConnect,
+  onSkip,
   isUpdate,
 }: {
   policies: ParsedSessionPolicies;
-  onConnect: (transaction_hash?: string, expiresAt?: bigint) => void;
+  onConnect: () => void;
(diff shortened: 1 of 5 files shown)


edisontim (approved):
lgtm!

packages/keychain/src/components/connect/CreateSession.tsx
```diff
@@ -85,21 +93,6 @@ const CreateSessionLayout = ({
...
   }, [controller, policies, maxFee, expiresAt, onConnect]);
 
-  const onSkipSession = useCallback(async () => {
-    if (!controller || !policies) return;
-    try {
-      setError(undefined);
-      setIsConnecting(true);
-
-      const processedPolicies = processPolicies(policies, true);
-      await controller.createSession(duration, processedPolicies, maxFee);
```
tarrencev, line 95: This is actually the intended logic, since we always create a wildcard session in this case and enforce policy logic locally. In this case, we approve no policies so the user is always prompted to sign. We should improve the function naming to reflect that better.

broody: Ah I missed the `toggleOff` input, will add back with better naming

PR #1844, opened 2025-07-22
fix: wait for account to be deployed to set controller

edisontim:


packages/keychain/src/components/connect/create/webauthn/index.ts
```diff
@@ -62,6 +63,8 @@ export function useWebauthnAuthentication() {
         },
       );
 
+      await waitUntilAccountIsDeployed(controller);
```
tarrencev, line 66: Hmm will this increase the users loading time?

edisontim: Only by a little, it's what I was telling you about with the session creation signature verification that is failing for some people, if they try to create the session too quickly the account contract isn't found

bal7hazar (commented):
Few comments, and was wondering if we cannot make something like this instead:

```ts
const endTime = Date.now() + timeout;
let isOwner = false;
while (Date.now() < endTime && !isOwner) {
  try {
    isOwner = await controller.provider.callContract(
      call,
      BlockTag.PENDING,
     );
  } catch (error) {
    ...
  }
}
```

packages/keychain/src/components/connect/create/webauthn/index.ts
tarrencev: yes but i think the better solution in that case is to do the simulation approach on the backend we discussed

PR #1934, opened 2025-08-26
feat: add lazyload option to defer iframe mounting

broody:
Add optional lazyload parameter to ControllerOptions that defers keychain iframe mounting until connect() is called. This reduces initial load time and prevents unnecessary resource fetching when the controller is instantiated but not immediately used.

- Add lazyload option to KeychainOptions interface
- Make IFrames.keychain optional to support lazy initialization
- Extract createKeychainIframe() method for reusable iframe creation
- Update connect() and probe() to ensure iframe exists before use
- Add null safety checks in async callbacks

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/controller/src/controller.ts
+++ b/packages/controller/src/controller.ts
@@ -53,19 +53,11 @@ export default class ControllerProvider extends BaseProvider {
 
     this.selectedChain = defaultChainId;
     this.chains = new Map<ChainId, Chain>();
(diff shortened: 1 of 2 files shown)


tarrencev (approved):

kronosapiens:
Thanks @broody !

PR #1990, opened 2025-09-09
fix: resolve inventory assets disappearing after tab navigation

tarrencev:
## Summary

Fixes a race condition in the `useTokens` hook where assets would disappear when navigating between tabs in the keychain inventory. Previously, only credits would remain visible after switching to another tab and back to inventory.

## Root Cause

The issue was caused by a race condition where:
1. When navigating between tabs, the component remounts and `toriiStatus` temporarily becomes "loading"
2. The original logic was filtering out token addresses during loading state (`if (toriiStatus === "loading") return []`)
3. This caused RPC data fetching to be skipped, leaving only credits (which are fetched separately) visible

## Changes

- **Remove toriiStatus dependency** from contractAddress filtering logic to prevent empty arrays during loading
- **Improve status determination** to consider both torii and RPC data sources instead of only relying on toriiStatus
- **Prevent empty state flickering** during navigation by defaulting to "success" when tokens are available

## Test plan

- [ ] Navigate to inventory tab and verify assets are displayed correctly
- [ ] Navigate to achievements/leaderboard/activity tabs
- [ ] Navigate back to inventory tab and verify all assets (not just credits) are still displayed
- [ ] Repeat navigation multiple times to ensure consistency

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/packages/keychain/src/hooks/token.ts
+++ b/packages/keychain/src/hooks/token.ts
@@ -295,15 +295,23 @@ export function useTokens(accountAddress?: string): UseTokensResponse {
     return newData;
   }, [rpcData, toriiTokens, countervalues]);
 
-  // Determine combined status - only show loading if both sources are loading
+  // Determine combined status based on actual data availability
   const combinedStatus = useMemo(() => {
-    // If we have tokens from either source, consider it success
-    if (tokens.tokens.length > 0) return "success";
-    // Only show loading if torii is actively loading
-    if (toriiStatus === "loading") return "loading";
-    // Default to success to prevent empty state flickering
+    // If torii is still loading on initial mount, show loading
+    if (
+      toriiStatus === "loading" &&
+      toriiTokens.length === 0 &&
+      rpcData.length === 0
+    ) {
+      return "loading";
+    }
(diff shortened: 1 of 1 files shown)


tarrencev:
## Additional Fix: Prevent Connect Controller Flash

I've also fixed the issue where the "Connect Controller" screen was briefly flashing when navigating between tabs.

### Root Cause
The flash occurred because when components remount during navigation, the controller state would momentarily be `undefined` before being re-initialized from storage. The `Home` component was immediately showing the Connect Controller screen for any falsy controller value.

### Solution
Added an `isControllerInitialized` state to distinguish between:
- Still checking for authentication (show loading screen)
- Confirmed no authentication (show Connect Controller)

This prevents the authentication screen from flashing during navigation while still showing it when genuinely needed.

### Changes in this commit:
1. Added `isControllerInitialized` state to the connection hook
2. Show loading screen while checking for stored controller
3. Only show Connect Controller after confirming no authentication exists
4. Updated test mocks to include the new state

Both issues (inventory assets disappearing and Connect Controller flash) are now resolved in this PR.

PR #2040, opened 2025-09-18
feat: removes old layout style override

nrmnqdds:


clicksave:
separate issue, but want to highlight it here. 

image


our button footers are missing top padding. 

image

--- 

https://linear.app/cartridge/issue/C7E-955/add-padding-top-to-the-controller-button-bar

clicksave:
Other than that, looking good!

PR #2059, opened 2025-09-24
Add support for WalletWithStarknetFeatures

kronosapiens:
This PR adds support for the `get-starknet` wallet standard by allowed the Controller to conform to the `WalletWithStarknetFeatures` interface.

## Context

@fracek reached out to us via Telegram asking us to add support for the `wallet-standard` interface (see [these docs](https://get-starknet.vercel.app/) and [this PR](https://github.com/starknet-io/get-starknet/pull/301) for more context) by creating a connector which conforms to the `WalletWithStarknetFeatures` interface.

The idea is to that [the `wallet-standard` interface](https://github.com/wallet-standard/wallet-standard) provides a higher-level interface than the current `StarknetWindowObject`, designed to be usable by frontend libraries like `starknet-react`, but also by command-line tools and more generic libraries like `solid.js`.

The `wallet-standard` interface calls for only four functions to be implemented, with return values being similar but not identical to those used by `StarknetWindowObject`: 

- StandardConnect (~`connect`)
- StandardDisconnect (~`disconnect`)
- StandardEvents (~`on`)
- StarknetWalletApi (~`request`)

## Approach 1

The first approach is to wrap the Controller in a custom connecter which implements `WalletWithStarknetFeatures` specifically for Controller. This implementation is available in `controllerWithStarknetFeatures.ts`.

UPDATE: this implementation has been removed from the PR.

## Approach 2

As an alternative to this connector, we can cast the Controller as follows:

```
import Controller from "@cartridge/controller";
import { StarknetInjectedWallet } from "@starknet-io/get-starknet-wallet-standard";

const controller = new StarknetInjectedWallet(new Controller())
```

The [`StarknetInjectedWallet` type](https://github.com/fracek/get-starknet/blob/develop/packages/wallet-standard/src/injected-wallet.ts#L27) implements `WalletWithStarknetFeatures` and wraps a `StarknetWindowObject`, which is what Controller implements:

```
export class StarknetInjectedWallet implements WalletWithStarknetFeatures {
  ...

  constructor(private readonly injected: StarknetWindowObject) {
   ...
}
 ```

To simplify the experience for developers, we expose an `asWalletStandard()` type on ControllerConnector which performs the cast and returns the correct type:

```
(new ControllerConnector()).asWalletStandard()
```

This approach minimizes code changes on our end, but does reduce the ability to customize the integration.

**I would personally lean towards this option for now.**

UPDATE: this is the approach we have chosen

## Approach 3

 Alternatively, we could direct developers to make the cast directly. This avoids any code changes on our end, although it does require documenting the typecast and limits our ability to customize the integration.

---

> [!NOTE]
> Adds `asWalletStandard()` to `ControllerConnector` using `StarknetInjectedWallet`, updates example import, expands TS libs, and adds a new wallet-standard dependency.
> 
> - **Connector**:
>   - Adds `ControllerConnector.asWalletStandard()` returning `WalletWithStarknetFeatures` via `StarknetInjectedWallet` in `packages/connector/src/controller.ts`.
>   - Introduces dependency `@starknet-io/get-starknet-wallet-standard@5.0.0-beta.0` in `packages/connector/package.json`.
> - **Examples**:
>   - Updates import to use `{ ControllerConnector }` from `@cartridge/connector` in `examples/next/src/components/DelegateAccount.tsx`.
> - **TypeScript**:
>   - Expands `lib` targets (adds `ES2017+`, `DOM`, `DOM.Iterable`) in `packages/connector/tsconfig.json`.
> - **Tooling**:
>   - Ignores `*.tgz` in `.gitignore`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 11dab3eaf9871e4ebb99704f1ee832ca63c3e047. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/packages/connector/package.json
+++ b/packages/connector/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@cartridge/connector",
   "version": "0.10.7",
-  "description": "Cartridge Controler Connector",
+  "description": "Cartridge Controller Connector",
   "module": "dist/index.js",
   "types": "dist/index.d.ts",
   "type": "module",
--- a/packages/connector/src/controller.ts
+++ b/packages/connector/src/controller.ts
@@ -45,6 +45,13 @@ export default class ControllerConnector extends InjectedConnector {
   }
 
   asWalletStandard(): WalletWithStarknetFeatures {
+    if (typeof window !== "undefined") {
+      console.warn(
+        `Casting Controller to WalletWithStarknetFeatures is an experimental feature and may contain bugs. ` +
+          `Please report any issues at https://github.com/cartridge-gg/controller/issues`,
+      );
+    }
+
     return new StarknetInjectedWallet(this.controller);
   }
 }
--- a/packages/connector/tsconfig.json
+++ b/packages/connector/tsconfig.json
@@ -2,6 +2,15 @@
   "$schema": "https://json.schemastore.org/tsconfig",
   "extends": "../tsconfig/base.json",
   "compilerOptions": {
+    "lib": [
+      "ES2017",
+      "ES2018",
+      "ES2019",
+      "ES2020",
+      "ESNext",
+      "DOM",
+      "DOM.Iterable"
+    ],
     "baseUrl": ".",
     "rootDir": "./src",
     "outDir": "./dist",

### CODE REVIEW

File: packages/connector/src/controllerWithStarknetFeatures.ts [hunk 1]
```diff
@@ -0,0 +1,124 @@
...
+      [StandardDisconnect]: {
+        version: "1.0.0" as const,
+        disconnect: this.controller.disconnect.bind(
+          this.controller,
+        ) as StandardDisconnectMethod,
+      },
+      [StandardEvents]: {
+        version: "1.0.0" as const,
+        on: this.controller.on.bind(this.controller) as StandardEventsOnMethod,
+      },
```
fracek, line 105: Does the `on` method of the controller match the `on` method of the wallet standard spec?

You can find the definition [here](https://github.com/wallet-standard/wallet-standard/blob/master/packages/core/features/src/events.ts).

Maybe you can use the object returned by `asWalletStandard` to avoid doing the translation yourself.

kronosapiens: Hmm, on closer look it seems like they are different specifications.

The existing `on` [implementation in BaseProvider](https://github.com/cartridge-gg/controller/blob/main/packages/controller/src/provider.ts#L176) has this signature:

```
  on: WalletEventListener = <E extends keyof WalletEventHandlers>(
    event: E,
    handler: WalletEventHandlers[E],
  ): void
```

With `WalletEventHandlers` defined in `starknet-io/types` [here](https://github.com/starknet-io/types-js/blob/aa42849cc87b83bfaef5ef5d1de73a88a280847a/src/wallet-api/events.ts#L7):

```
export type AccountChangeEventHandler = (accounts?: string[]) => void;
export type NetworkChangeEventHandler = (chainId?: ChainId, accounts?: string[]) => void;

export interface WalletEventHandlers {
  accountsChanged: AccountChangeEventHandler;
  networkChanged: NetworkChangeEventHandler;
}
```

Whereas the [implementation required by the `wallet-standard` `WalletWithStarknetFeatures`](https://github.com/wallet-standard/wallet-standard/blob/master/packages/core/features/src/events.ts#L61C1-L64C17) is this:

```
export type StandardEventsOnMethod = <E extends StandardEventsNames>(
    event: E,
    listener: StandardEventsListeners[E]
) => () => void;
```

With the `StandardEventsListeners` [defined in the same file in](https://github.com/wallet-standard/wallet-standard/blob/master/packages/core/features/src/events.ts#L77C1-L85C2) `wallet-standard` as:

```
export interface StandardEventsListeners {
    change(properties: StandardEventsChangeProperties): void;
}
```

This is making me wonder whether adapting Controller to work with `wallet-standard` won't require a re-write of the whole Controller around a different interface.

fracek: Now that I look at this PR again, I think this class is not really needed. The `asWalletStandard()` method on the controller is enough to integrate with libraries that target get-starknet. As long as the starknet window object you return has the correct metadata (name, icon, etc) it will look correct.

kronosapiens: Ok. I don't want to spend too much more time going down this rabbit hole for now, so what I'm thinking is to ship the `asWalletStandard()` function with a console warning that this is an experimental feature, and to share a link to log any issues. We'll see how it goes and take it from there.

kronosapiens:
@broody can you take a look at this and approve if you like it?

broody (approved):
LGTM!

PR #2064, opened 2025-09-25
feat: create controller with username autocomplete

nrmnqdds:


clicksave:
@nrmnqdds 

Nice work. Noticing a few things with the latest preview though. 

---

image

image

[] Sometimes I get a "create new" rendering for controllers that already exist

----

image

[] Exact matching results should appear at the top of the list

PR #2152, opened 2025-10-29
fix: removes skeleton and blank screen on create controller

nrmnqdds:
Resolves C7E-1137

- [x] skeleton page
- [x] blank screen flash for no policies
- [x] blank screen flash for verified policies

### DIFF
--- a/packages/keychain/src/components/connect/create/useCreateController.ts
+++ b/packages/keychain/src/components/connect/create/useCreateController.ts
@@ -63,26 +63,53 @@ const createSession = async ({
   policies,
   params,
   handleCompletion,
+  closeModal,
+  searchParams,
 }: {
   controller: Controller;
   policies?: ParsedSessionPolicies;
   params?: ReturnType<typeof parseConnectParams>;
   handleCompletion: () => void;
+  closeModal?: () => void;
+  searchParams: URLSearchParams;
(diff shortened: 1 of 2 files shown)


packages/keychain/src/components/ConnectRoute.tsx
```diff
@@ -52,56 +52,20 @@ export function ConnectRoute() {
...
-        }
-      };
-
-      void createSessionForVerifiedPolicies();
-    }
-  }, [params, controller, policies, handleCompletion]);
+  // Don't render anything if we don't have controller yet - CreateController handles loading
+  if (!controller) {
+    return null;
+  }
```
tarrencev: Can we remove?

packages/keychain/src/components/connect/create/useCreateController.ts
```diff
@@ -218,6 +241,52 @@ export function useCreateController({
...
+            cleanupCallbacks(params.params.id);
+            handleCompletion();
+          } else {
+            void closeModal?.();
+          }
+          return;
+        }
+
+        // Handle verified policies - create session automatically and close modal
+        if (policies.verified) {
```
tarrencev: seems this logic is duplicated below. can we extract common parts to a helper?

PR #2355, opened 2026-01-27
Add Capacitor session example

Larkooo:
Adds a Capacitor-based session example and wires it into the workspace. Updates session handling to normalize redirect payloads and align executeFromOutside behavior across runtimes. Includes docs and Node example tweaks for consistency.


.tool-versions
```diff
@@ -1,3 +1,3 @@
 scarb 2.9.4
 starknet-foundry 0.38.3
-nodejs v20.11.1
+nodejs 22.22.0
```
broody: Can this remain node v20? A lot of the CI is pin to v20 version and also vercel.

PR #2401, opened 2026-02-11
fix: SessionProvider preset support and address normalization

### LINKED ISSUE #2399: SessionProvider: missing preset support and address normalization causing session/not-registered errors
## Summary

`SessionProvider` (the redirect-based flow for environments where iframes don't work, like Capacitor/mobile) has several gaps compared to `ControllerProvider` that are causing `session/not-registered` errors and developer confusion.

There are two related but distinct problems:

1. **SessionProvider does not support presets** — developers must manually duplicate preset policies
2. **Address normalization mismatch** between the keychain and SDK sides of the session flow can cause hash divergence

### Current behavior

`SessionProvider` (`packages/controller/src/session/provider.ts`) accepts a `policies` field but has no `preset` option:

```typescript
export type SessionOptions = {
  rpc: string;
  chainId: string;
  policies: SessionPolicies;  // required, no preset alternative
  redirectUrl: string;
  // ...
};
```

When `SessionProvider` constructs the keychain redirect URL (line 262-268), it serializes `policies` directly into the URL — it never sends a `preset` parameter.

Meanwhile, the keychain's `useConnectionValue` hook (`packages/keychain/src/hooks/connection.ts:468-486`) does support loading presets from the URL. So the infrastructure is there — `SessionProvider` just doesn't use it.

### Impact

- Developers who configure presets at https://github.com/cartridge-gg/presets expect them to work with `SessionProvider`, but they don't.
- Developers must manually hardcode policies in their `SessionProvider` initialization, duplicating what's already in their preset.
- When a preset is updated, `SessionProvider` users keep using stale policies — a drift that can silently cause `session/not-registered` errors.
- This has been independently confirmed by a developer porting a game to Capacitor/mobile, who had to build a custom `NativeProvider` to work around the gap.

### Related: other missing ControllerProvider features

`SessionProvider` also lacks `openProfile()`, `openStarterPack()`, `openSettings()`, and `openPurchaseCredits()`, which all depend on the iframe model. This is a broader feature gap for mobile/Capacitor developers.

### Root cause

When a session is created and later used, the policies are hashed into a merkle root via `toWasmPolicies()` (`packages/controller/src/utils.ts:100-159`). For the hash to match, both sides must produce identical input to `toWasmPolicies`.

The **keychain side** (session creation) normalizes addresses via `getChecksumAddress` during policy parsing:
- `parseUrlPolicies` calls `toSessionPolicies`, which calls `getChecksumAddress` on contract addresses
- `parseSessionPolicies` further processes contracts (adds metadata, re-sorts by ERC20 type)

The **SDK side** (session usage in `SessionProvider.tryRetrieveFromQueryOrStorage()`) passes `this._policies` to `toWasmPolicies` where `this._policies` was built directly from the developer's raw input — **without address normalization**.

`toWasmPolicies` sorts canonically by address (case-insensitive) before hashing, which was the fix in #2357. However, the `target` field passed to the WASM layer uses the address as-is. If the WASM layer includes the target bytes in the hash computation, different casings (e.g. `0x049d36...` vs `0x049D36...`) would produce different hashes.

### How this manifests

1. Developer passes policies with a non-checksummed address to `SessionProvider`
2. Keychain creates the session using checksummed addresses (via `getChecksumAddress`)
3. `SessionProvider` reconstructs the session using the original non-checksummed addresses
4. The WASM-computed session hash diverges → `session/not-registered`

### For preset support

**Option A: Add `preset` to `SessionOptions`** (minimal change)

Add an optional `preset` field to `SessionOptions` and include it in the redirect URL:

```typescript
export type SessionOptions = {
  rpc: string;
  chainId: string;
  policies?: SessionPolicies;  // make optional when preset is provided
  preset?: string;
  redirectUrl: string;
  // ...
};
```

When `preset` is provided, append `?preset=<name>` to the keychain URL. The keychain already handles this — `useConnectionValue` reads `preset` from URL params and calls `loadConfig()`.

**Option B: Resolve presets on the SDK side**

Have `SessionProvider` call `loadConfig(preset)` itself, resolve the chain-specific policies, and use those as the canonical policies. This keeps the SDK as the single source of truth.

### For address normalization

Normalize addresses in the `SessionProvider` constructor using `getChecksumAddress` before storing `this._policies`:

```typescript
this._policies = {
  verified: false,
  contracts: policies.contracts
    ? Object.fromEntries(
        Object.entries(policies.contracts).map(([address, contract]) => [
          getChecksumAddress(address),  // normalize here
          { ...contract, methods: ... }
        ])
      )
    : undefined,
  messages: ...
};
```

This should also be done in the SDK-side `parsePolicies` function (`packages/controller/src/policies.ts`) for consistency.

Alternatively, `toWasmPolicies` itself could normalize addresses before passing them to WASM, which would fix the issue for all callers.

## Files involved

| File | Role |
|------|------|
| `packages/controller/src/session/provider.ts` | SessionProvider — no preset support, no address normalization |
| `packages/controller/src/policies.ts` | SDK-side `parsePolicies` — no address normalization |
| `packages/controller/src/utils.ts` | `toWasmPolicies` — canonical sort but no address normalization |
| `packages/controller/src/iframe/keychain.ts:96-107` | Policy/preset precedence logic for ControllerProvider |
| `packages/keychain/src/hooks/connection.ts:468-486` | Preset loading (already supports presets from URL) |
| `packages/keychain/src/hooks/connection.ts:630-664` | Policy resolution (preset vs URL policies) |
| `packages/keychain/src/hooks/session.ts:77-148` | `parseSessionPolicies` — normalizes addresses, re-sorts by type |

## Related issues

- #2357 — canonical policy sorting fix in `toWasmPolicies`

kronosapiens:
## Summary

Fixes #2399.
Addresses `session/not-registered` errors caused by policy hash divergence between the SDK and keychain.

### SDK changes (`@cartridge/controller`)

- **Add `preset` and `shouldOverridePresetPolicies` to `SessionOptions`**: Developers can now pass `preset: "my-game"` instead of manually duplicating policies.
  Preset policies are resolved via `loadConfig()` from `@cartridge/presets`, matching the keychain's behavior.
  Policy precedence matches `ControllerProvider`: preset wins unless `shouldOverridePresetPolicies: true`.
- **Consolidate async init into `_init()`**: Preset resolution, session retrieval from storage, and signer key setup all happen in a single eagerly-started async method.
  Public methods (`username`, `probe`, `connect`) simply `await this._ready`.
- **Normalize addresses in `toWasmPolicies`**: Applies `getChecksumAddress()` to contract targets before hashing, preventing divergence from inconsistent address casing.
- **Extract `getPresetSessionPolicies` utility**: Shared between `SessionProvider` and the keychain's `getConfigChainPolicies`, replacing ~25 lines of defensive type narrowing with a clean helper.
- **Warn on silent policy drop**: When both `preset` and `policies` are provided without the override flag, a `console.warn` is logged (previously policies were silently ignored in `ControllerProvider`).
- **Send `preset` in keychain URL**: When preset is used, `connect()` sends `&preset=` instead of `&policies=`, so the keychain resolves the same policies from the same source.

### Keychain changes

- **Gate rendering on `isPoliciesResolved` in `Authentication`**: The `Authentication` wrapper now waits for preset policies to fully resolve before rendering child routes.
  This replaces scattered `isPoliciesResolved` guards that were previously in `ConnectRoute` and eliminates race conditions where components render before async preset policy resolution completes.

### Root cause

When a developer configured presets and used `SessionProvider`, the keychain would resolve policies from the preset while the SDK used the developer's manually-passed policies.
If these differed (or if addresses had different casing), the session hash would diverge, producing `session/not-registered` errors.

### Design note

Preset resolution is async (`loadConfig` fetches from the network), but JS constructors can't be async.
The `_init()` method is eagerly started in the constructor and stores its promise as `_ready`.
A cleaner alternative would be a static factory method (`SessionProvider.create(options)`), but that's a breaking API change — worth considering in a future major version.

### Follow-up

- Apply same preset support to `TelegramProvider` and Node.js `SessionProvider`

## Test plan

- [x] All 45 controller unit tests pass (including updated `toWasmPolicies` tests)
- [x] All 356 keychain tests pass
- [x] `tsc --noEmit` passes with zero errors
- [x] Full monorepo build succeeds
- [x] Lint and format checks pass
- [ ] Manual verification with a preset-configured app using SessionProvider

### DIFF
--- a/packages/controller/src/iframe/keychain.ts
+++ b/packages/controller/src/iframe/keychain.ts
@@ -106,7 +106,7 @@ export class KeychainIFrame extends IFrame<Keychain> {
       _url.searchParams.set("preset", preset);
       if (policies) {
         console.warn(
-          "[Controller] Both `preset` and `policies` provided. " +
+          "[Controller] Both `preset` and `policies` provided to ControllerProvider. " +
             "Policies are ignored when preset is set. " +
             "Use `shouldOverridePresetPolicies: true` to override.",
         );
--- a/packages/controller/src/session/provider.ts
+++ b/packages/controller/src/session/provider.ts
@@ -4,7 +4,7 @@ import {
   signerToGuid,
   subscribeCreateSession,
 } from "@cartridge/controller-wasm";
-import { loadConfig } from "@cartridge/presets";
+import { loadConfig, SessionPolicies } from "@cartridge/presets";
 import { AddStarknetChainParameters } from "@starknet-io/types-js";
 import { encode } from "starknet";
 import { API_URL, KEYCHAIN_URL } from "../constants";
@@ -53,8 +53,8 @@ export default class SessionProvider extends BaseProvider {
   private _ready: Promise<void>;
   protected _keychainUrl: string;
   protected _apiUrl: string;
-  protected _publicKey: string;
-  protected _sessionKeyGuid: string;
+  protected _publicKey!: string;
+  protected _sessionKeyGuid!: string;
   protected _signupOptions?: AuthOptions;
   public reopenBrowser: boolean = true;
 
@@ -82,7 +82,6 @@ export default class SessionProvider extends BaseProvider {
     // 3. Otherwise, use provided policies
     if ((!preset || shouldOverridePresetPolicies) && policies) {
       this._policies = parsePolicies(policies);
-      this._preset = undefined;
     } else {
       this._preset = preset;
       if (policies) {
@@ -249,28 +248,17 @@ export default class SessionProvider extends BaseProvider {
 
   async username() {
     await this._ready;
-    this.tryRetrieveFromQueryOrStorage();
     return this._username;
   }
 
   async probe(): Promise<WalletAccount | undefined> {
-    if (this.account) {
-      return this.account;
-    }
-
     await this._ready;
-    this.account = this.tryRetrieveFromQueryOrStorage();
     return this.account;
   }
 
   async connect(): Promise<WalletAccount | undefined> {
-    if (this.account) {
-      return this.account;
-    }
-
     await this._ready;
 
-    this.account = this.tryRetrieveFromQueryOrStorage();
     if (this.account) {
       return this.account;
     }
--- a/packages/controller/src/utils.ts
+++ b/packages/controller/src/utils.ts
@@ -105,9 +105,10 @@ export function toSessionPolicies(policies: Policies): SessionPolicies {
 /**
  * Converts parsed session policies to WASM-compatible Policy objects.
  *
- * IMPORTANT: Policies are sorted canonically before hashing. Without this,
- * Object.keys/entries reordering can cause identical policies to produce
- * different merkle roots, leading to "session/not-registered" errors.
+ * IMPORTANT: Policies are sorted canonically and addresses are normalized
+ * via getChecksumAddress before hashing. Without this, Object.keys/entries
+ * reordering or inconsistent address casing can cause identical policies to
+ * produce different merkle roots, leading to "session/not-registered" errors.
  * See: https://github.com/cartridge-gg/controller/issues/2357
  */
 export function toWasmPolicies(policies: ParsedSessionPolicies): Policy[] {


broody (approved):
Works great!

PR #2581, opened 2026-04-29
chore(payment): remove coinbase postMessage debug logs

broody:
## Summary

Removes the `[coinbase-popup]` and `[coinbase-hook]` `console.log` instrumentation that was added while debugging the postMessage relay between the Coinbase popup and the keychain. These logs were noisy in production (every postMessage event, every poll, every popup-close watcher tick) and no longer needed.

Kept the `console.error` / `console.warn` calls for genuine edge cases:
- `window.opener is null` (relay can't happen)
- "No Coinbase events received after iframe load" (server-error fallback)
- Ignored-after-success guards
- Non-terminal `commit_error` and `not supported` warnings

Also picks up regenerated GraphQL types for paymaster STRK fees (same as #2580).

## Test plan

- [ ] Coinbase Apple Pay flow still completes end-to-end
- [ ] Cancelling/closing the Coinbase popup behaves as before
- [ ] Browser console is quiet during a normal purchase

### DIFF
--- a/packages/keychain/src/components/coinbase-popup.tsx
+++ b/packages/keychain/src/components/coinbase-popup.tsx
@@ -91,9 +91,7 @@ export function CoinbasePopup() {
     const handleKeychainMessage = (event: MessageEvent) => {
       if (event.origin !== keychainOrigin) return;
       if (event.data?.__coinbase_relay) return; // Ignore our own relayed events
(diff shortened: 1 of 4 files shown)


broody:
Folding into #2580 so the Apple Pay rollout and the Coinbase log cleanup ship together.

PR #2585, opened 2026-05-08
fix(keychain): send turnkey otp public key proof

tarrencev:
## Summary
- encrypt SMS OTP codes with @turnkey/crypto before verify
- split SMS auth into verifyOtp V2 followed by otpLogin V2
- send compressed P-256 targetPublicKey and raw-hex clientSignature for Turnkey login

## Review
- Scope check: CLEAN. The diff is limited to the Turnkey SMS key/signature path and related state plumbing.
- Pre-landing review: no blocking findings found. The client uses encryptOtpCodeToBundle and converts the iframe DER stamp to raw 64-byte hex before sending.

## Tests
- pnpm --filter @cartridge/keychain... build
- pnpm --filter @cartridge/keychain test:ci -- src/components/connect/create

## Notes
- Full pre-commit previously failed in unrelated @cartridge/controller Jest suites because @starknet-io/get-starknet-core could not be resolved in this checkout.


tarrencev:
Checked Turnkey docs and updated the patch accordingly.

Relevant docs:
- SMS docs describe the upgraded flow as `INIT_OTP_V3` -> `VERIFY_OTP_V2` -> `OTP_LOGIN_V2`, with `encryptedOtpBundle` generated from `otpEncryptionTargetBundle`.
- The OTP migration guide says `encryptOtpCodeToBundle` from `@turnkey/crypto` is for that upgraded V2 verification flow.
- For our current iframe/legacy flow, the SDK path that produces compressed P-256 hex uses `@turnkey/encoding` `pointEncode`; this PR now uses that helper directly instead of adding a direct `@turnkey/crypto` dependency.

This PR is intentionally only the key-format fix for the current API. A full migration to Turnkey's upgraded OTP flow would require coordinated backend changes to `VERIFY_OTP_V2` / `OTP_LOGIN_V2`.

tarrencev:
Updated this PR with the SMS OTP signing fix.

The frontend now generates a separate OTP verification P-256 keypair, uses that public key in `encryptedOtpBundle`, and signs the `OTP_LOGIN_V2` client signature with the matching private key. The iframe public key remains the session/credential-bundle target.

This depends on internal PR https://github.com/cartridge-gg/internal/pull/4661 because prod API currently rejects `clientSignature.publicKey != targetPublicKey`.

rsodre:
Implemented in #2593

PR #2621, opened 2026-06-17
fix(keychain): verify origin in popup auth (/auth) flow

starknetdev:
## Problem

On **iOS** (every browser — all WebKit, incl. iOS Chrome), signing into a Controller preset navigates to the `x.cartridge.gg/auth` popup flow (introduced in #2583 for Safari, because WebAuthn + cookies can't run in a cross-origin iframe under ITP). On that path users get a spurious **"This application is not verified"** warning **and** are forced to manually approve session policies — even though the policies are identical to the working desktop/Android version and the app origin is whitelisted in the preset. Android/desktop Chrome stay in the iframe and are unaffected.

## Root cause

The popup `/auth` flow conveys the app origin via the `origin` query param and sends **no `redirect_url`**. `computeVerifiedState()` (standalone/`!isIframe()` branch) only verified `redirect_url`, so it returned `false` for popup auth regardless of the preset. That single `verified=false` flag drives both symptoms: the warning, and (via `requiresSessionApproval`) the forced policy approval.

## Why not just trust the `origin` param

A first cut verified the `origin` query param — but that's **caller-supplied and spoofable**, and it breaks the property the redirect flow relied on: trust must be bound to the **delivery target**. An attacker could pass `origin=<trusted game>` with `redirect_url=<attacker>` and get an auto-approved session delivered to themselves. So that approach was dropped.

## Fix: root verification in sources the caller cannot forge

- **Redirect flow** → verify `redirect_url` (the keychain navigates back to it, so the claimed origin and the recipient are the same — can't be pointed elsewhere).
- **Popup `/auth` flow** → the opener (keychain iframe, which holds the **penpal-enforced** parent origin) pushes that origin to the popup over a **same-origin-validated `postMessage` handshake** (`auth-ready` → `auth-context`). The popup no longer derives its verified origin from the URL. This rides the same opener↔popup channel #2583 already uses to deliver the session, so it's as reliable as the shipping completion path.

`computeVerifiedState` now delegates to `verifyStandaloneOrigin(currentOrigin, redirectUrl, allowedOrigins)`, and the standalone connection effect no longer reads the spoofable `origin` param for the origin state. The URL `origin` param is retained only for the pre-existing WebAuthn appId / username-prefill use (not verification).

## Why the handshake can't be spoofed

The popup accepts an `auth-context` **only if** `event.origin === x.cartridge.gg` **and** `event.source === window.opener` **and** the `channelId` matches.

| Attack | Why it fails |
|---|---|
| Open `/auth?…&origin=<trusted>` directly from `evil.com` | Popup posts `auth-ready` to `window.opener` with `targetOrigin=x.cartridge.gg`; an `evil.com` opener never receives it. A guessed `auth-context` has `event.origin=evil.com` → rejected. The result `postMessage` is likewise locked to `x.cartridge.gg`, so `evil.com` can't receive the session either. |
| Embed the **real** keychain iframe in `evil.com` | The iframe's origin from penpal is `evil.com` (browser-enforced); it forwards *that* → popup correctly reads **unverified**. It cannot be made to report a different origin. |
| Inject / race an `auth-context` from another window | Fails `event.source === window.opener` (identity) and/or `event.origin === x.cartridge.gg`. |

Assumptions (all standard, none new): browser-enforced `postMessage` `targetOrigin`/`event.origin`/`event.source`; the popup's `window.opener` is the keychain iframe (already relied on by #2583); penpal's parent origin is browser-enforced (the same origin already gating all parent↔iframe RPC); no XSS on `x.cartridge.gg`. If `window.opener` is ever null (COOP), the popup **fails safe** — warning shown, manual approval required — rather than failing open.

## Scope note

This affects **iOS only** (WebKit) — not game-specific, and not Android/desktop. The `origin` URL param is left in place for the pre-existing appId path; happy to harden that separately if desired.

## Tests

- `verifyStandaloneOrigin` — incl. a **spoof regression**: `origin=<trusted>` + `redirect_url=<attacker>` stays unverified.
- `parseTrustedAuthContext` — cross-origin sender / wrong source / wrong channel / bad payload all rejected.

Full `connection.test.ts` + new `popup.test.ts` green; eslint, prettier, and `tsc -b` clean.

### DIFF
--- a/packages/keychain/src/hooks/connection.test.ts
+++ b/packages/keychain/src/hooks/connection.test.ts
@@ -1,10 +1,10 @@
 import {
   getStandaloneAppOrigin,
   getStandaloneRedirectUrl,
-  getStandaloneVerificationOrigin,
   isNestedIframe,
   isOriginVerified,
   resolvePolicies,
+  verifyStandaloneOrigin,
 } from "./connection";
 import { vi } from "vitest";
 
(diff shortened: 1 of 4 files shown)


starknetdev:
Reworked per the spoofing feedback — verification no longer trusts the caller-supplied `origin` query param.

- **Redirect flow** verifies the delivery target (`redirect_url`), so a claimed origin can't diverge from where the session is actually delivered.
- **Popup `/auth` flow** gets the **penpal-enforced** parent origin from the opener (keychain iframe) over a same-origin-validated `postMessage` handshake — rooted in the same browser-enforced origin penpal already uses to gate parent↔iframe RPC. It rides the same opener↔popup channel #2583 already uses to deliver the session.

Added a regression test for the `origin`-vs-`redirect_url` mismatch (stays unverified) and tests for the handshake validator (cross-origin sender / wrong source / wrong channel rejected). The full threat walk-through is in the updated PR description.

The URL `origin` param is left only for the pre-existing WebAuthn appId path (not verification) — can address that separately if you'd like.

PR #2634, opened 2026-07-14
feat: upgrade Controller to Starknet.js v10

tarrencev:
Upgrades Controller, keychain, connectors, and examples to Starknet.js `10.0.2`, React 19, Node 22, published `@cartridge/arcade@0.4.0`/Dojo 2, and wallet-standard discovery through `@starknet-start/*`.

Preserves the hosted keychain protocol for Controller `0.13.12` and `0.13.13`, adds serialized authoritative chain switching, and keeps the exact Starknet override required by published Starknet Start 1.x metadata.

Adds real Controller/Session Starknet Start integration coverage, an 8/8 cross-version browser contract matrix, and packed npm/pnpm/Bun consumers that enforce one physical Starknet `10.0.2` installation.

This is the breaking `0.14.0` migration; release remains gated behind a separately reviewed `0.14.0-alpha.1` prerelease.


examples/next/src/components/providers/StarknetProvider.tsx
```diff
@@ -338,7 +336,6 @@ export function StarknetProvider({ children }: PropsWithChildren) {
       autoConnect
       defaultChainId={BigInt(defaultChainId)}
       chains={starknetConfigChains}
-      connectors={[controllerConnector, session]}
```
tarrencev, line 349: Hmm do we still need this?

examples/next/src/components/InvalidTxn.tsx
```diff
@@ -55,19 +57,16 @@ export function InvalidTxn() {
...
           onClick={() =>
-            account.execute(
-              [
-                {
-                  contractAddress: ETH_CONTRACT_ADDRESS,
-                  entrypoint: "approve",
-                  calldata: [account.address, "0x0", "0x0"],
-                },
-              ],
-              { maxFee: 1000000000000000000n },
```
tarrencev, line 66: Revert?

packages/keychain/src/components/funding/Deposit.tsx
```diff
@@ -190,22 +197,26 @@ function DepositInner({ onComplete }: DepositProps) {
                 <div className="flex flex-col items-center gap-4">
                   <div className="w-full flex gap-4">
                     {connectors
-                      .filter((c) => ["argentX", "braavos"].includes(c.id))
+                      .filter((c) =>
+                        ["argent", "braavos"].some((name) =>
```
tarrencev: hmm intended? should be "ready" now?

examples/next/src/components/providers/StarknetProvider.tsx
rsodre: correct, it doesn't accept or need the connectors anymore
in the example, it appears that `useConnect()` finds it in `window.starknet_connector`

rsodre:
Neither the controller or session connectors in the next.js and minimal examples are reconnecting automatically on reload.

The Starknet provider has its `autoConnect` feature disabled. I opened an issue to understand why.
https://github.com/starknet-start/starknet-start/issues/597

This is a client issue, when using `@starknet-start/react` 1.0.8, should not be a blocker

PR #2654, opened 2026-07-17
fix: avoid purchase flash after location verification

tarrencev:
Keeps the location gate mounted while the resumed asynchronous purchase initializes, preventing the review page from flashing between verification and Coinflow. The pending gate now clears only after the continuation settles and safely clears on errors. Adds a regression test that holds the continuation open and verifies the gate remains visible. Full pre-commit lint, formatting, tests, build, and GraphQL generation passed.


tarrencev:
Closing after reproducing the actual host-side behavior in Glitch Bomb. The apparent refresh is Glitch Bomb's global loading screen reactivating while account-scoped data hydrates after Controller connect; this controller-only transition patch does not address that root cause.

PR #2658, opened 2026-07-21
fix(keychain): disable text selection globally

bal7hazar:
## Summary

- Apply `select-none` on `html, body` in the keychain so UI text (usernames, labels, tooltips, etc.) is not selectable.
- Re-enable selection with `select-text` on `input` and `textarea` so typing/editing in form fields keeps working.

## Testing

- Verified on the local dev server: computed styles show `user-select: none` on `html`/`body` and `user-select: text` on inputs, with no console errors.

### DIFF
--- a/packages/keychain/src/index.css
+++ b/packages/keychain/src/index.css
@@ -11,7 +11,12 @@
 @layer base {
   html,
   body {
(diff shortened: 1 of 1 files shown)


bal7hazar:
Closing — not needed for now.

